Issue #42, opened 2013-12-22
Data serialization - Data format {feature-request}


ORNis:
Recent commits add some additional data formats. At the time being there are multiple formats in openMVG
- Camera files (one for simple pinhole camera model, another brown camera model)
- Image list files
- Camera database files

My suggestions to improve the situation goes in two directions
- Factorize all existing "formats". For example coupling the list of image files camera parameters files in a kind of global openMVG format
- Adopt a "standard" serialization format (JSON, XML...)

My preference goes to JSON because it's lightweight and flexible and in that sense it meets the requirements of openMVG. Moreover it offer a very good interoperability with other languages and potential services built on top of openMVG.

The major drawback of inclusion of a standard serialization format is that it could add another dependency to a third party library. In that sense it will lead to a less maintainable codebase. Another possibility should be to only adopt the JSON as a structure (instead of comma separated text) and perform I/O without an additional library.

Some light and modern JSON libraries
https://github.com/hjiang/jsonxx
https://github.com/sirikata/json-spirit (but depends on boost spirit...)
https://github.com/anhero/JsonBox
the three seem to be under MIT License.

ORNis:
I think this could be close now...

pmoulon:
You're right! @ORNis 
Now JSON/XML/BINARY input/output formats are natively supported thanks to the great cereal library.

PR #77, opened 2014-02-06
Switch to JSON for file formats

Meekohi:
See issue #42. Adds JsonBox to the project for reading/writing JSON files (trust me, I tried several libraries and this is the one you want). So far only switched lists.txt to imageParams.json, but more to follow.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,21 @@
+language: cpp
+
+compiler:
+  - gcc
+
+install: sudo apt-get install cmake libpng-dev libjpeg8-dev libxxf86vm1 libxxf86vm1 libxxf86vm-dev x11proto-xf86vidmode-dev libxrandr-dev
+
+before_script:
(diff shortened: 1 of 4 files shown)


fabiencastan:
Any plan to merge that?

simogasp:
+1

pmoulon:
Integrate directly this input format in the 0.8 is a bit dangerous, but will could definitively consider it as an I/O module for the 0.9.

I would prefer first add a binary to export some data in JSON for the 0.8 and smoothly go to a full JSON module I/O for the 0.9.

We could for example write first a main_openMVG2JSON that would convert an openMVG SfM scene in a set of JSON files.

What do you think about that?

fabiencastan:
I would prefer to avoid this intermediate step that will create more confusion.
Most of those files are internal files, so it should not have a big impact on users. It should just be clear in the release note.

PR #214, opened 2015-01-08
WIP : generate one matches file per image (for parallelization)

eprana:
* Add main_extractFeatures
* Add PairBuilder to generate pairs from one image
* One matches file per image

### DIFF
--- a/src/openMVG/matching/indMatch_utils.hpp
+++ b/src/openMVG/matching/indMatch_utils.hpp
@@ -9,6 +9,7 @@
 #define OPENMVG_MATCHING_IND_MATCH_UTILS_H
 
 #include "openMVG/matching/indMatch.hpp"
+#include "third_party/stlplus3/filesystemSimplified/file_system.hpp"
 #include <map>
 #include <fstream>
 #include <iterator>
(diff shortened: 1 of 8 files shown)


pmoulon:
I think we must discuss since all those pull request are realized a bit too soon.
I think it's better to finish the 0.8 version and add those new features in the 0.9 version.

eprana:
Yes, these PR should not be merged yet (we rename the title to WIP to make it clear). But we open it for discussion.

The current state allows us to parallelize detection and matching on renderfarm.

The next step is to remove all indexes from file formats (to use image names). That will allow to add new images to the reconstruction afterwards.

pmoulon:
Hi,
Thanks @eprana .

I plan also to extend to support 'full image names':
That will bring:
- support to image that have the same basename but that are in different directory,
- but brings other challenges, what happens when you delete, add new images...
Images must have a unique ID, and images must always increase.

Regarding parallelism I'm more to keep the existing main_computeMatches (it supports already an optional pair list file).
So your master machine can compute some pair list to create a job for each pc of your cluster without too much modifying the current binary (just the output matches.X.txt produced file names must be an option, to do not have conflicts).
Once all your matches___.X.txt will be computed, you can merge them and continue to run the actual SfM pipeline.

We will continue the discussion to see how to manage that in the long term with a nice design.
Don't forgot that some BOW matching (Bag of Words) can also be a good tip to enhance openMVG scalability.

fabiencastan:
1/ Unique key
We really need to be able to move images without breaking things, so the full path isn't a good option.
We would like to the filename as the unique key everywhere. If we have multiple photos from different camera with the same filename, we could easily add a prefix before using openMVG.
We can even add a script to check that and rename automatically if we have input images from multiple paths.

2/ matches.X.txt
Is there a reason to merge them in one file instead of using each file?

3/ BOW
Yes,  sure!

> Regarding parallelism I'm more to keep the existing main_computeMatches (it supports already an optional pair list file).

Not sure to understand the point.

fabiencastan:
Will be done in another PR with the new IOs.

PR #220, opened 2015-01-22
Added check for _bHtmlReport in destructor to avoid seg faults

alexanderkoumis:
Added check for _bHtmlReport flag in IncrementalSfMEngine destructor to avoid the segmentation fault that occurs in case the flag is set to false.


pmoulon:
Thank, fixed also in the GlobalReconstructionEngine.

PR #271, opened 2015-05-02
OptimizeForArchitecture.cmake uses gcc's -march=native option if available

daeyun:
I noticed that this script doesn't recognize my Haswell CPU and sets the target architecture as "core2". Some of what it does is not really necessary anymore because GCC 4.2>= has the -march=native option. This patch replaces AutodetectHostArchitecture() with -march=native when it's available.


pmoulon:
@daeyun Thanks.

PR #310, opened 2015-05-25
Fixes for MinGW compilation, detecting some corner cases

rhiestan:
Fixes for MinGW compilation
Detecting corner cases: Checking for empty matches
Using std::move where applicable

### DIFF
--- a/src/openMVG/matching_image_collection/Matcher_Regions_AllInMemory.cpp
+++ b/src/openMVG/matching_image_collection/Matcher_Regions_AllInMemory.cpp
@@ -98,8 +98,9 @@ void Template_Matcher(
     const features::Regions *regionsI = regions_perImage.at(I).get();
     const size_t regions_countI = regionsI->RegionCount();
     const std::vector<PointFeature> pointFeaturesI = regionsI->GetRegionsPositions();
-    const typename MatcherT::ScalarT * tabI =
-      reinterpret_cast<const typename MatcherT::ScalarT *>(regionsI->DescriptorRawData());
+    const typename MatcherT::ScalarT * tabI = NULL;
+    if(regions_countI > 0)
+      tabI = reinterpret_cast<const typename MatcherT::ScalarT *>(regionsI->DescriptorRawData());
 
     MatcherT matcher10;
     ( matcher10.Build(tabI, regions_countI, regionsI->DescriptorLength()) );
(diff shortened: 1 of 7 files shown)


src/openMVG/matching_image_collection/Matcher_Regions_AllInMemory.cpp
```diff
@@ -98,8 +98,9 @@ void Template_Matcher(
     const features::Regions *regionsI = regions_perImage.at(I).get();
     const size_t regions_countI = regionsI->RegionCount();
     const std::vector<PointFeature> pointFeaturesI = regionsI->GetRegionsPositions();
-    const typename MatcherT::ScalarT * tabI =
-      reinterpret_cast<const typename MatcherT::ScalarT *>(regionsI->DescriptorRawData());
+    const typename MatcherT::ScalarT * tabI = NULL;
+    if(regions_countI > 0)
```
pmoulon: While not consider
```
const typename MatcherT::ScalarT * tabI = NULL;
if(regions_countI == 0)
  continue;
```

src/openMVG/matching_image_collection/Matcher_Regions_AllInMemory.cpp
```diff
@@ -112,8 +113,9 @@ void Template_Matcher(
 
       const features::Regions *regionsJ = regions_perImage.at(J).get();
       const size_t regions_countJ = regionsJ->RegionCount();
-      const typename MatcherT::ScalarT * tabJ =
-        reinterpret_cast<const typename MatcherT::ScalarT *>(regionsJ->DescriptorRawData());
+      const typename MatcherT::ScalarT * tabJ = NULL;
+      if(regions_countJ > 0)
```
pmoulon: While not consider
```
const typename MatcherT::ScalarT * tabJ = NULL;
if(regions_countJ == 0)
  continue;
```

src/openMVG/matching_image_collection/Matcher_Regions_AllInMemory.cpp
```diff
@@ -149,7 +151,7 @@ void Template_Matcher(
         ++my_progress_bar;
         if (!vec_FilteredMatches.empty())
         {
-          map_PutativesMatches.insert( make_pair( make_pair(I,J), vec_FilteredMatches ));
+          map_PutativesMatches.insert( make_pair( make_pair(I,J), std::move(vec_FilteredMatches) ));
```
pmoulon: I should consider the move instruction in other place:
https://github.com/openMVG/openMVG/blob/develop/src/openMVG/matching_image_collection/Matcher_Regions_AllInMemory.cpp#L152

src/openMVG/matching_image_collection/Matcher_Regions_AllInMemory.cpp
rhiestan: Yes, you are right, this is better.

src/openMVG/matching_image_collection/Matcher_Regions_AllInMemory.cpp
rhiestan: Where do you mean? The link points to the change I already made.

pmoulon: You're right, sorry.

pmoulon:
@rhiestan Merged.
Thanks

PR #313, opened 2015-05-28
Add support for MVS Texturing

ORNis:


igorti:
@ORNis Since mvs-texturing is expecting a mesh, not a point cloud that openMVG creates - what utility is supposed to be used along with this tool to create mesh? MVE's `fssrecon` doesn't seem to be working out of the box with openMVG's point cloud.

pmoulon:
I think that @ORNis made this tool to deliver valid data to mvs-texturing and he certainly used a mesh done with another tool.

ORNis:
yes! you can prefer, for example, using MVS-Texturing over TextureMesh in openMVS (even if they are both based on the same paper) or imagine a pipeline like this: openMVG + PMVS + MVS-Texturing etc... I can agree it's less relevent now that there is an openMVG -> NVM converter and an openMVG -> MVE one because MVS-texturing also "knows" these two formats  but my tiny exporter is still way faster  (it do es not export keypoints).

igorti:
Thanks for quick answers! My intention is to create mesh + texture from the sparse point cloud directly since I'm interested mainly in planar scenes and the only pipeline that works out of the box seems to be openMVG -> openMVS. I guess it should be possible to tweak `fssrecon` so that it can create mesh from openMVG point cloud as well, then, together with mvs-texturing, it would probably be a good alternative.

ORNis:
I think the best way to deal with meshing of sparse point clouds should be something like this https://arxiv.org/abs/1604.06258 (and related publication from Maxime l'Huillier) but the simplest way to do something with fssr would be
- Compute normals (with PCA)
- Orient normals correctly thanks to visibility information 
- Use a constant scale value for each point (or maybe something better but for now I do not see anything)
the main issue would be that the sparse point cloud could be highly anisotropic and thus your normals could be wrong or noisy because classic normal estimation methods use a fixed kernel size. so maybe it's better to use something like this https://sites.google.com/site/boulchalexandre/publications/sgp2012_normals.

igorti:
Thanks for the suggestion! Unfortunately, I have limited knowledge in this area to implement this, but if you might be interested, please drop me an email and I'll gladly discuss further.

PR #331, opened 2015-06-25
Fixed problem with binary json files

rhiestan:
Files are now opened with std::ios::binary to avoid end-of-line
conversion in binary files on Windows.

### DIFF
--- a/src/openMVG/sfm/sfm_data_io_cereal.hpp
+++ b/src/openMVG/sfm/sfm_data_io_cereal.hpp
@@ -44,7 +44,7 @@ bool Load_Cereal(
   const bool b_structure = (flags_part & STRUCTURE) == STRUCTURE;
 
   //Create the stream and check it is ok
-  std::ifstream stream(filename.c_str());
+  std::ifstream stream(filename.c_str(), std::ios::binary | std::ios::in);
   if (!stream.is_open())
     return false;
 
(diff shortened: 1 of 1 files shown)


rhiestan:
This change fixes the problem with binary cereal files on Windows, at least in my tests.
The drawback of this change is that newly created JSON files now have Unix end-of-lines on all platforms, also on Windows.

pmoulon:
What about consider open file in binary mode only if we want save binary files?
So only the binary format will be impacted.

rhiestan:
I also thought about it. However, I don't see a drawback in JSON files being in Unix format (except that Notepad will not display the files properly).

pmoulon:
Does the compatibility is kept with the old json file?
I will also consider to add a conversion (read-save) unit test in sfm_data_io_test since this unit test is missing

rhiestan:
Yes, I can read an "old" JSON file with openMVG_main_ConvertSfM_DataFormat and save it to binary, then convert back to JSON. The file is identical except for the line endings and doubles (at the 13th/14th digit). Also XML files are identical except for the line endings, and they can be read with old and new versions of openMVG_main_ConvertSfM_DataFormat.
Binary files are not compatible, however.

pmoulon:
Tested on windows and JSON and XML looks fine with Notepad ;-)
:+1 @rhiestan

PR #346, opened 2015-07-28
Add shared library support. More deps can now be external.

cbalint13:
This PR does:

* Add **OpenMVG_BUILD_SHARED** build option (OFF by default).
* Add **.so versioning** schema in case of shared library.
* Subfolders "dependencies" & "third_party" **should** be compiled with -fPIC and STATIC.
* FLANN, COINUTILS, CLP, OSI, LEMON can now be **external** ones (optional using DIR_HINTS).
* Centralized folder for the **output build** e.g. 'Linux-x86_64-Release' (makes 'make tests' happy).
* **Display** appropriate settings and found library & versions at end of cmake session.
* Minor fixes (some broken naming conventions e.g. for main_exifParsing)

I hope this work will enjoy good time for good reasons.


cbalint13:
**Sample from Fedora 22 using all external libraries:**


 # /usr/bin/cmake -DCMAKE_C_FLAGS_RELEASE:STRING=-DNDEBUG -DCMAKE_CXX_FLAGS_RELEASE:STRING=-DNDEBUG -DCMAKE_Fortran_FLAGS_RELEASE:STRING=-DNDEBUG -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON -DCMAKE_INSTALL_PREFIX:PATH=/usr -DINCLUDE_INSTALL_DIR:PATH=/usr/include -DLIB_INSTALL_DIR:PATH=/usr/lib64 -DSYSCONF_INSTALL_DIR:PATH=/etc -DSHARE_INSTALL_PREFIX:PATH=/usr/share -DLIB_SUFFIX=64 -DBUILD_SHARED_LIBS:BOOL=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo -DOpenMVG_BUILD_SHARED=ON -DCMAKE_SKIP_RPATH=ON -DOpenMVG_BUILD_TESTS=ON -DOpenMVG_BUILD_EXAMPLES=ON -DOpenMVG_BUILD_COVERAGE=OFF -DOpenMVG_BUILD_OPENGL_EXAMPLES=OFF -DEIGEN_INCLUDE_DIR_HINTS=/usr/include/eigen3 -DCERES_DIR_HINTS=/usr/include/ceres/ -DFLANN_DIR_HINTS=/usr/include/flann/ -DCLP_DIR_HINTS=/usr/include/coin/ -DOSI_DIR_HINTS=/usr/include/coin/ -DCOINUTILS_DIR_HINTS=/usr/include/coin/ -DLEMON_DIR_HINTS=/usr/include/lemon/ ../src/

-----/////-------

** OpenMVG version: 0.8.1
** Build Shared libs: ON
** Build OpenMVG tests: ON
** Build OpenMVG samples applications: ON
** Build OpenMVG openGL examples: OFF
** Enable code coverage generation: OFF
** Enable OpenMP parallelization: ON
** Build OpenCV+OpenMVG samples programs: OFF
** Use OpenCV SIFT features: OFF


-- EIGEN: 3.2.5 (external)
-- CERES: 1.10.0 (external)
-- FLANN: 1.8.4 (external)
-- LIBTIFF: 4.0.3 (external)
-- LIBPNG: 1.6.16 (external)
-- LIBPNG (external)
-- CLP: 1.16.6 (external)
-- COINUTILS: 2.10.8 (external)
-- OSI: 0.107.4 (external)
-- LEMON: 1.3.1 (external)


-- Configuring done
-- Generating done

src/openMVG/graph/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
 
-UNIT_TEST(openMVG_graph connectedComponent "lemon")
-UNIT_TEST(openMVG_graph triplet_finder "lemon")
+UNIT_TEST(openMVG_graph connectedComponent "")
```
pmoulon: strange that lemon is not longer required...

cbalint13: @pmoulon ,

I notice that, BOOST replaced it but traces are still present, please check:
find src/ -type f -exec grep -H 'lemon/' {} \; | grep -v third_party

If agree, to be 'correct' i leave LEMON for now just like it was, i believe we should clean up LEMON in another PR.

--------------<<<------------
Lets see some examples:

A) BOOST replaces just fine:
src/openMVG/multiview/rotation_averaging_l1.cpp

B) boost does not replaces at all:
src/openMVG/graph/graph.hpp
-------------->>>--------------

src/CMakeLists.txt
```diff
@@ -246,17 +230,151 @@ ADD_DEFINITIONS(-DEIGEN_MPL2_ONLY)
...
-ENDIF(Ceres_FOUND)
+ENDIF()
+
+# ==============================================================================
+# Flann
+# ==============================================================================
+# - internal by default (flann),
+# - external if FLANN_DIR_HINTS and a valid Flann setup is found
+# ==============================================================================
+FIND_PATH(FLANN_INCLUDE_DIRS flann.hpp ${FLANN_DIR_HINTS})
```
simogasp: I'am not an expert but isn't it better to pass through a `find_package()` instead of looking for the `.h` of each library inside the `LIBNAME_DIR_HINTS`? It seems to me that this can work only if the library that you are looking for are installed in the usual system directories (eg /usr/lib): when u later call `FIND_LIBRARY(LIBNAME_LIBRARY NAMES libname)` you don't provide any other directory in which to look for. If the library is installed in a non system path there's no way it could look into it.
For example, if have installed flann in some directory in my home `/home/simone/my/personal/flann/install/path/` then according to this version I should pass `FLANN_DIR_HINTS=/home/simone/my/personal/flann/install/path/include/flann` (where the `flann.hpp` is installed) but then the `FIND_LIBRARY(FLANN_LIBRARY NAMES flann_cpp)` will have no clue that the .so are placed somewhere inside `/home/simone/my/personal/flann/install/path/lib`.

pmoulon: @simogasp @cbalint13 
This is a first draft, perhaps some Find_packageX can be added in src/cmakeFindModules and make this CMakeLists.txt file more clear.

cbalint13: @simogasp , @pmoulon 

* There is no native find_package() for clp, coinutils, osi, lemon, flann.

* Yes LIBNAME_DIR_HINTS tell /hints where headers are. Here i could use additional PATH_SUFFIXES (that can lookup all standard system include paths too) but then i am afraid it will find _always_ find the header files, so better give it by _exact_ way or use internal one.

* For the ".so" FIND_LIBRARY() will found them by looking up all your standard  $LD_LIBRARY_PATH on the system. Yes he has "clues" so no need to specify an exact location for libs too. If libs are not in LD_LIBRARY_PATH than there is a terrible situation not only finding the libs but then using them by runtime.

-----<<<---

Now that these are more clear, let me further propose variants:

1. Leave it just how is it now with find_library() & find_path() tandem.
2. Do "Find_packageX.cmake" for each (will relay on proposal no1 indeed).
3. Use special cmake's FindPkgConfig & Pkg_Check_Modules(Clp clp).

-----<<<----

* For No.3 proposal Clp, CoinUtils, Osi and Lemon seems to have the 'pkg-config' style thing. 

$ rpm -ql coin-or-Clp-devel | grep .pc
/usr/lib64/pkgconfig/clp.pc
/usr/lib64/pkgconfig/osi-clp.pc

* Now, a lookup for such package would look like few lines:

// require pkg-config tool.
INCLUDE(FindPkgConfig REQUIRED)
// search Clp
IF(PKG_CONFIG_FOUND)
    INCLUDE(FindPkgConfig)
    PKG_CHECK_MODULES(Clp clp)
ENDIF()

* After these lines ${CLP_INCLUDE_DIRS} #{CLP_LIBRARIES} are ready to consume.

I'll vote No.3 if do not mine to bring additional pkg-config rool dependency here, and with this i believe we consume all ideas, at least on my side.

Thanks !

simogasp: @cbalint13 Thanks, I've learnt something new about `FIND_LIBRARY()` :-)
I think 2) for the lib having some ''official'' `FindLibname.cmake` and 3) for the others should be fine. Or, at least, it makes sense to me :-)

pmoulon: I think FLANN is exporting also a .pc file.
Do you know if FindPkgConfig is working directly on windows or if something have to be configured?

cbalint13: @simogasp , @pmoulon 

I agree with Simone,

* Lets have single FindLibname.cmake as an option for others (that does not have even pkg-config ways), perhaps a nice parametrizable & reusable generic one.

* In our case of Clp, Osi, ClpUtils, Lemon and even for Flann as i've checked now all have official pkg-config handlers so we could exploit cmake internal pkg-config lookup routines.

Lets see Pierre what would do, if he agrees to i refactor this PR asap.

pmoulon: Let's have FindX.cmake,
I think it's more clean and make new FindX.cmake file that could be used by the community.
Those cmake file can use FindPkgConfig or basic FIND_LIBRARY code (depending of the user needs).
BTW, the VERSION number display for each 3rd party library was a really good idea.

I don't know if Clp, Osi, CoinUtils should be merge in only one FINDX or if we need check for the three and deal with perhaps a missing one...

cbalint13: @pmoulon ,
* True, flann too seems to have .pc file.
* Clp, Osi, CoinUtils has .pc too lets exploit them too if agree !

I also wandering if we would need FINDX.cmake for each or an universal one, i leave the decision to you.

cbalint13: @pmoulon ,

  An now i realize that going pkg-config way will miss the nice VERSION feature. So lets better agree with intial No.2 proposal and do one FindX.cmake for each one library since VERSION can be extracted quite different ways for each lib so universal FindX.cmake will be not feasable at some point.
 
 I now start to see clear advantages for having one FindX.cmake for each lib, and deal them in slight different way.

src/CMakeLists.txt
```diff
@@ -361,6 +468,83 @@ ELSE (EXISTS ${SPHINX_EXECUTABLE})
...
+ENDIF()
+
+IF(DEFINED OpenMVG_USE_INTERNAL_PNG)
+  MESSAGE(STATUS "LIBPNG: " ${PNG_VERSION_STRING} " (internal)")
+ELSE()
+  MESSAGE(STATUS "LIBPNG: " ${PNG_VERSION_STRING} " (external)")
+ENDIF()
+
+IF(DEFINED OpenMVG_USE_INTERNAL_JPEG)
+  MESSAGE(STATUS "LIBPNG (internal)")
```
pmoulon: must be LIBJPEG

src/CMakeLists.txt
```diff
@@ -361,6 +468,83 @@ ELSE (EXISTS ${SPHINX_EXECUTABLE})
...
+ELSE()
+  MESSAGE(STATUS "LIBPNG (external)")
```
pmoulon: must be LIBJPEG

src/CMakeLists.txt
cbalint13: OK,
 
  * And since these has been argued let me present the FindX.cmake (one per lib) way, allow me some hours of work first, then will see our further issues with linkage.

src/CMakeLists.txt
```diff
@@ -246,17 +230,151 @@ ADD_DEFINITIONS(-DEIGEN_MPL2_ONLY)
...
+  SET(LEMON_INCLUDE_DIRS
+     ${CMAKE_CURRENT_SOURCE_DIR}/third_party/lemon
+     ${PROJECT_BINARY_DIR}/third_party/lemon)
+  SET(LEMON_LIBRARY lemon)
+ENDIF()
+
+# ==============================================================================
+# Internal CLP/OSI/COINUTILS libraries:
+# ==============================================================================
+ADD_SUBDIRECTORY(dependencies/osi_clp/)
```
pmoulon: what about consider adding it only in the case of internal clp and co.
```
IF(DEFINED OpenMVG_USE_INTERNAL_CLP AND DEFINED OpenMVG_USE_INTERNAL_OSI)
  ADD_SUBDIRECTORY(dependencies/osi_clp/)
ENDIF()
```

src/CMakeLists.txt
```diff
@@ -136,46 +148,18 @@ IF (OpenMVG_BUILD_OPENGL_EXAMPLES)
...
+  SET(LP_INCLUDE_DIRS
     ${MOSEK_INCLUDE}
     ${CMAKE_CURRENT_SOURCE_DIR}/dependencies/osi_clp/Osi/src/OsiMsk/
   )
   INCLUDE_DIRECTORIES(${LP_INCLUDE_DIRS})
 ENDIF(MOSEK_FOUND)
 
-#- osi_clp (linear programming interface)
-# ==============================================================================
-if(NOT EXISTS ${PROJECT_SOURCE_DIR}/dependencies/osi_clp/Clp)
```
pmoulon: I would love to keep this submodule check.
It helps beginner to ensure they have a valid OpenMVG local repository that is ready for compilation.
What about just change the submodule that is tested, since CLP can now be external...
```
# ==============================================================================
# Check that submodule have been initialized and updated
# ==============================================================================
if(NOT EXISTS ${PROJECT_SOURCE_DIR}/dependencies/cereal/include)
  message(FATAL_ERROR
    "\n submodule(s) are missing, please update your repository:\n"
    "  > git submodule update -i\n")
endif()
```

src/CMakeLists.txt
pmoulon: Don't worry about timing issue.
I was just mentioning all my thoughts.

src/CMakeLists.txt
cbalint13: The LIBJPEG things clearly are typos.

pmoulon: :8ball: it happens often and code review helps to spot them :+1:

src/CMakeLists.txt
cbalint13: Hmm, github cannot patch recursive sub-projects ?!

This one below (screwed due to github) seems to be lost ! Should i open separate PR for openMVG-ThirdParty ?

---------------->>>>---------------
diff -Nru openMVG/src/dependencies/osi_clp/CMakeLists.txt openMVG-shared/src/dependencies/osi_clp/CMakeLists.txt
--- openMVG/src/dependencies/osi_clp/CMakeLists.txt     2015-07-23 02:33:27.000000000 +0300
+++ openMVG-shared/src/dependencies/osi_clp/CMakeLists.txt      2015-07-28 22:56:56.060799915 +0300
@@ -5,6 +5,15 @@
 # CMake version control
 CMAKE_MINIMUM_REQUIRED(VERSION 2.6)
 
+
+# static libraries only
+SET(BUILD_SHARED_LIBS OFF)
+
+# use PIC code for link into shared lib
+IF(UNIX)
+  SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC")
+ENDIF(UNIX)
+

 INCLUDE_DIRECTORIES(.
   ./CoinUtils/src/
@@ -28,6 +37,20 @@
   ) 
 ENDIF(MSVC)
 
-ADD_SUBDIRECTORY(Osi)
-ADD_SUBDIRECTORY(Clp)
-ADD_SUBDIRECTORY(CoinUtils)
+
+IF(DEFINED OpenMVG_USE_INTERNAL_COINUTILS)
+  ADD_SUBDIRECTORY(CoinUtils)
+  SET_PROPERTY(TARGET lib_CoinUtils PROPERTY FOLDER OpenMVG/3rdParty/osi_clp)
+ENDIF()
+
+IF(DEFINED OpenMVG_USE_INTERNAL_OSI)
+  ADD_SUBDIRECTORY(Osi)
+  SET_PROPERTY(TARGET lib_Osi PROPERTY FOLDER OpenMVG/3rdParty/osi_clp)
+ENDIF()
+
+IF(DEFINED OpenMVG_USE_INTERNAL_CLP)
+  ADD_SUBDIRECTORY(Clp)
+  SET_PROPERTY(TARGET lib_OsiClpSolver PROPERTY FOLDER OpenMVG/3rdParty/osi_clp)
+  SET_PROPERTY(TARGET lib_clp PROPERTY FOLDER OpenMVG/3rdParty/osi_clp)
+ENDIF()
+
----------------<<<<---------------

pmoulon: Sorry about that, yes please fell free to do a PR for https://github.com/openMVG-thirdparty/osi_clp.
Once it is done, we will be able to update the commit id in the OpenMVG repository to make the submodule match the last osi_clp modifications.

cbalint13: > Sorry about that, yes please fell free to do a PR for
> https://github.com/openMVG-thirdparty/osi_clp.
> Once it is done, we will be able to update the commit id in the OpenMVG
> repository to make the submodule match the last osi_clp modifications.
>
>
>
I'll do that as solution now, in fact we have this
http://openrisc.rdsor.ro/openMVG-deps.patch . I cannot enforce this upper
level due to local fPIC / STATIC thing.

cbalint13:
* Reworked this PR, sqashed all into https://github.com/cbalint13/openMVG/commit/15db54c65b419cca61ca9163e9b440f347727be2
* This PR against osi_clp **must** be also merged: https://github.com/openMVG-thirdparty/osi_clp/pull/2

pmoulon:
@cbalint13 Can you check that I have done the right job with your PR (I have added some modification on my side on top of it)?
We must also add a BUILD page to the online doc in order to explain the build modes:
- internal,
- external librairies.

Thanks a lot for your efforts over this PR.

PR #350, opened 2015-08-01
Added docker for building openMVG docker images

FreakTheMighty:
I think it would be great to build an official openMVG docker image. I think it would encourage people to experiment with openMVG by making the install process easier. Also, this could be a good base for someone interesting in building a REST API, web interface or cli for openMVG.

I haven't experimented with the image very much yet, so I'm sure there will be additional changes needed, but this should be a good starting place. After pulling the image, the sfm software executables can be run like so:

`docker run freakthemighty/openmvg openMVG_main_ComputeFeatures`

Check it out on docker hub:

https://registry.hub.docker.com/u/freakthemighty/openmvg/

If this does get merged, it would be great if openMVG could create their own docker hub account so that build triggers can be setup. A new image can be created automatically after each commit and for multiple branches. This would make it very easy for people to try out, and switch between, bleeding edge builds and stable builds.


pmoulon:
Hi @FreakTheMighty 
Thank for your docker script.
You're right it could be a very neat idea to test it under a server-client environment or to test the build across various configuration.

Regarding the continuous integration, by now Travis-ci is used but docker can be added also to test other build configuration.

Thanks for the Pull request, it is now noted as a feature for the upcoming 0.9 version.

FreakTheMighty:
Hey @pmoulon 

I'm not thinking about docker as a replacement for travis. The main use case is making it easy for people to use openMVG's command line applications. I hope this docker image can make it easy for people with docker installed to run openMVG with just two commands.

pmoulon:
Can you check that you can use a more recent GCC on docker (like 4.8)?

FreakTheMighty:
Will do. It shouldn't be a problem. I'll update the Dockefile and let you know.

pmoulon:
Thanks ;-)

FreakTheMighty:
A new build is in progress https://hub.docker.com/r/freakthemighty/openmvg/builds/

pmoulon:
Since the last PR the BUILD path is not longer /opt/openMVG_Build/software/SfM but set to a directory with a variable name: SET(EXECUTABLE_OUTPUT_PATH "${PROJECT_BINARY_DIR}/${CMAKE_SYSTEM_NAME}-${CMAKE_SYSTEM_PROCESSOR}-${CMAKE_BUILD_TYPE}")

In order to fix that we can consider adding a specific install path
``̀`
-DCMAKE_INSTALL_PREFIX="/opt/openMVG_Build/install"
# and modify
ENV PATH $PATH:/opt/openMVG_Build/software/SfM
# to become
ENV PATH $PATH:/opt/openMVG_Build/install/bin
```

Sorry for this change...

FreakTheMighty:
No problem, I can look into the change this evening.

FreakTheMighty:
Alright, I've made the changes you suggested. I also replaced the git clone with a simple `ADD`. This should ensure that dockerhub builds do in fact build the specific branch that triggered the build.

I need to do additional testing, the builds take a while on my laptop, so the one cooking right now isn't fully tested.

pmoulon:
Done: https://github.com/openMVG/openMVG/commit/12e908bf8a49fbe7fe3541a273a472b3a39613ce
Thanks

PR #358, opened 2015-08-11
If CLP/OSI/COINUTILS  external cmake stops.

cbalint13:
This small PR address case when CLP/OSI/COINUTILS is external. I believe the fix is obvious. It fixes such failures, when one tries to build with all external ones:

---------------<<<------------
-- Found Ceres version: 1.10.0 installed in: /usr
-- Flann 1.8.4 found (include: /usr/include/flann)
-- CoinUtils 2.10.8 found (include: /usr/include/coin)
-- Clp 1.16.6 found (include: /usr/include/coin)
-- Osi 0.107.4 found (include: /usr/include/coin)
CMake Error at CMakeLists.txt:325 (message):
  
   When you want to use OSI or CLP or COINUTILS as external librairies,
   please use the three as external.
 --------------->>>-------------


pmoulon:
Thanks. https://github.com/openMVG/openMVG/commit/f46cf39fb3ff098a23b786bce0d6c5506a37b4e4

PR #359, opened 2015-08-11
adding tangential distortion model for cameras

davidli43:
unit test passed for add_disto and remove_disto

### DIFF
--- a/src/openMVG/cameras/Camera_Pinhole_Brown.hpp
+++ b/src/openMVG/cameras/Camera_Pinhole_Brown.hpp
@@ -1,0 +1,144 @@
+// Author: Sida Li
+
+#ifndef OPENMVG_CAMERA_PINHOLE_BROWN_HPP
+#define OPENMVG_CAMERA_PINHOLE_BROWN_HPP
+
+#include "openMVG/numeric/numeric.h"
+#include "openMVG/cameras/Camera_Common.hpp"
+
+#include <vector>
+
+namespace openMVG {
+namespace cameras {
+
+/// Implement a Pinhole camera with a 3 radial distortion coefficients ad 2 tangential distortion coefficients.
+/// x_d = x_u (1 + K_1 r^2 + K_2 r^4 + K_3 r^6) + (T_2 (r^2 + 2 x_u^2) + 2 T_1 x_u y_u)
+/// y_d = y_u (1 + K_1 r^2 + K_2 r^4 + K_3 r^6) + (T_1 (r^2 + 2 y_u^2) + 2 T_2 x_u y_u)
+class Pinhole_Intrinsic_Brown_T2 : public Pinhole_Intrinsic
+{
+    protected:
+    // center of distortion is applied by the Intrinsics class
+    std::vector<double> _params; // K1, K2, K3, T1, T2
+
+    public:
+
+    Pinhole_Intrinsic_Brown_T2(
+        int w = 0, int h = 0,
+        double focal = 0.0, double ppx = 0, double ppy = 0,
+        double k1 = 0.0, double k2 = 0.0, double k3 = 0.0,
+        double t1 = 0.0, double t2 = 0.0)
+            :Pinhole_Intrinsic(w, h, focal, ppx, ppy)
+    {
+        _params.resize(5);
+        _params[0] = k1;
+        _params[1] = k2;
+        _params[2] = k3;
+        _params[3] = t1;
+        _params[4] = t2;
+    }
+
+    EINTRINSIC getType() const { return PINHOLE_CAMERA_TANGENTIAL2; }
+
+    virtual bool have_distro() const { return true;}
+
+    virtual Vec2 add_disto(const Vec2 & p) const{
+        return (p + distoFunction(_params, p));
+    }
+
+    // numerical approximation based on
+    // Heikkila J (2000) Geometric Camera Calibration Using Circular Control Points.
+    // IEEE Trans. Pattern Anal. Mach. Intell., 22:1066-1077
+
+    virtual Vec2 remove_disto(const Vec2 & p) const{
+        double epsilon = 1e-8; //criteria to stop the iteration
+        Vec2 p_u = p;
+        while((add_disto(p_u)-p).lpNorm<1>() > epsilon)//manhattan distance between the two points
+        {
+            p_u = p - distoFunction(_params, p_u);
+        }
+        return p_u;
+    }
+
+    // Data wrapper for non linear optimization (get data)
+    virtual std::vector<double> getParams() const
+    {
+        std::vector<double> params = Pinhole_Intrinsic::getParams();
+        params.push_back(_params[0]);
+        params.push_back(_params[1]);
+        params.push_back(_params[2]);
+        params.push_back(_params[3]);
+        params.push_back(_params[4]);
+        return params;
+    }
+
+    // Data wrapper for non linear optimization (update from data)
+    virtual bool updateFromParams(const std::vector<double> & params)
+    {
+        if (params.size() == 8) {
+          *this = Pinhole_Intrinsic_Brown_T2(
+            _w, _h,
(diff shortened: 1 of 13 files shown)


davidli43:
I needed tangential distortion for my project, so I wrote the class for tangential distortion.
The Brown–Conrady model is used to account for the tangential distortion.
The numerical approximation of removing distortion is supported by Heikkila J (2000).

Related publications are at:
https://eserv.asprs.org/PERS/1966journal/may/1966_may_444-462.pdf
http://www2.imm.dtu.dk/~jmca/02501/dum/image_rectification/Geometric%20Camera%20Calibration.pdf

src/openMVG/cameras/Camera_Pinhole_Brown.hpp
```diff
@@ -0,0 +1,144 @@
+// Author: Sida Li
```
pmoulon: You must add the licence and copyright in this way:
```
// Copyright (c) 2012, 2013 Sida Li.

// This Source Code Form is subject to the terms of the Mozilla Public
// License, v. 2.0. If a copy of the MPL was not distributed with this
// file, You can obtain one at http://mozilla.org/MPL/2.0/.
```

src/openMVG/cameras/Camera_Pinhole_Brown.hpp
```diff
@@ -0,0 +1,144 @@
...
+
+#include "openMVG/numeric/numeric.h"
+#include "openMVG/cameras/Camera_Common.hpp"
+
+#include <vector>
+
+namespace openMVG {
+namespace cameras {
+
+/// Implement a Pinhole camera with a 3 radial distortion coefficients ad 2 tangential distortion coefficients.
```
pmoulon: ad => and

src/openMVG/cameras/Camera_Pinhole_Brown.hpp
```diff
@@ -0,0 +1,144 @@
...
+    public:
+
+    Pinhole_Intrinsic_Brown_T2(
+        int w = 0, int h = 0,
+        double focal = 0.0, double ppx = 0, double ppy = 0,
+        double k1 = 0.0, double k2 = 0.0, double k3 = 0.0,
+        double t1 = 0.0, double t2 = 0.0)
+            :Pinhole_Intrinsic(w, h, focal, ppx, ppy)
+    {
+        _params.resize(5);
```
pmoulon: Tip:
you can do
```
_params = {k1, k2, k3, t1, t2};
```

src/openMVG/cameras/Camera_Pinhole_Brown.hpp
```diff
@@ -0,0 +1,144 @@
...
+        _params[0] = k1;
+        _params[1] = k2;
+        _params[2] = k3;
+        _params[3] = t1;
+        _params[4] = t2;
+    }
+
+    EINTRINSIC getType() const { return PINHOLE_CAMERA_TANGENTIAL2; }
+
+    virtual bool have_distro() const { return true;}
```
pmoulon: must be
```
have_disto
```

src/openMVG/cameras/Camera_Pinhole_Brown.hpp
```diff
@@ -0,0 +1,144 @@
...
+    virtual Vec2 add_disto(const Vec2 & p) const{
+        return (p + distoFunction(_params, p));
+    }
+
+    // numerical approximation based on
+    // Heikkila J (2000) Geometric Camera Calibration Using Circular Control Points.
+    // IEEE Trans. Pattern Anal. Mach. Intell., 22:1066-1077
+
+    virtual Vec2 remove_disto(const Vec2 & p) const{
+        double epsilon = 1e-8; //criteria to stop the iteration
```
pmoulon: const epsilon

src/openMVG/cameras/Camera_Pinhole_Brown_test.cpp
```diff
@@ -0,0 +1,41 @@
+// Author: Sida Li
```
pmoulon: You must add the licence and copyright in this way:
```
// Copyright (c) 2012, 2013 Sida Li.

// This Source Code Form is subject to the terms of the Mozilla Public
// License, v. 2.0. If a copy of the MPL was not distributed with this
// file, You can obtain one at http://mozilla.org/MPL/2.0/.
```

pmoulon:
Thanks @davidli43 I put some minor comments.
Thanks for the pull request.
I will have a brief look soon and I think it can be integrated for the next release.

We must just fix the name:
Since it's the implementation of the Brown COnrady model:
Why you don't have kept the predefined PINHOLE_CAMERA_BROWN?
=> I mean your implementation file is named Camera_Pinhole_Brown,so use Brown everywhere could be nice.

davidli43:
I have found a bug in my distoFunction formulation. The original (x,y) was not multiplied to the radial distortion. Found out about it while testing sequential bundle adjustment. 
I am integrating the Brown model into the SfM pipeline.
Still testing with data sets, I will push those changes when they are fully tested.

@pmoulon Is there anything that I should be aware of when integrating this new model?
Here's what I have done so far:
1. Added custom cost functor accounting for tangential distortion
2. Added new case for all the switch case control flows
3. Changing default camera model to PINHOLE_CAMERA_BROWN; I know this can also be done by passing in an extra parameter, but for testing purposes, i'm just going to change them for now.

I do have a question though, in the sequential sfm file https://github.com/Skycatch/openMVG/blob/master/src/openMVG/sfm/pipelines/sequential/sequential_SfM.cpp#L889
when bKnownIntrinsic is false, the intrinsics are set with the decomposition of the projection matrix from robustResection(). 
I'm not quite sure about the math behind KRt_From_P(), so I was wondering if you could point me to any resources that can help me figure out what values to put for t1 and t2 when creating the new camera intrinsic group?

davidli43:
@pmoulon Yes, I've integrated the model into full structure from motion pipeline in my development branch. My tests have shown significant reduction in drift, also the RMSE in the reconstruction report. The current pull request only contains the brown model class. It hasn't been integrated into the pipeline yet. I'll be updating the integrated files this week.

pmoulon:
Thanks, the remaining step normally consists in making the Camera model accessible from the ImageListing step. So add a new case in this switch https://github.com/openMVG/openMVG/blob/develop/src/software/SfM/main_SfMInit_ImageListing.cpp#L270 and extend the doc.

davidli43:
As for extending the doc, is this page the only one I need to edit? https://openmvg.readthedocs.org/en/latest/openMVG/cameras/cameras/
Just add an extra option for "OpenMVG Pinhole camera models" ?

pmoulon:
Please add an entry here:
https://github.com/openMVG/openMVG/blob/develop/docs/sphinx/rst/openMVG/cameras/cameras.rst#openmvg-pinhole-camera-models
=> three factor radial distortion + tangential distortion

And add an entry here:
https://github.com/openMVG/openMVG/blob/develop/src/software/SfM/main_SfMInit_ImageListing.cpp#L97

davidli43:
I found out that the camera sensor file was missing in the develop branch.
Not sure if this is intended but I added it back so that I can run reconstruction tests.

davidli43:
@pmoulon I've just ran the SfM_Sequential pipeline against two data sets that I have using the Brown Conrady Model. The results were great, with significant reduction in drift. Currently the pipeline still uses Radial 3 Model as default, with the Brown model as an option. I have completed all my tests and everything is good to go.

pmoulon:
Thanks @davidli43.
The feature is integrated.
See https://github.com/openMVG/openMVG/commit/32e8608aa3cca1f50e5d152cd60e7ef087a3a0f4 in which I have squashed your commits.
Tested on some dataset (tangential estimated factors was very small).
I think the effect will be more visible on images with large overlap like aerial sequences.

Notes: Adding back the camera database was not necessary, it moved to https://github.com/openMVG/openMVG/tree/develop/src/openMVG/exif/sensor_width_database

PR #369, opened 2015-08-19
Alembic exporter

cpichard:
Hi Pierre,

here is the code for simple alembic exporter. We use alembic internally as a pivot format for geometry and camera, so it would be awesome if we finally manage to get this format in openMVG. 
In any case, I can make the build optional, let me know !

Best,

Cyril

### DIFF
--- a/src/cmakeFindModules/FindAlembic.cmake
+++ b/src/cmakeFindModules/FindAlembic.cmake
@@ -18,7 +18,8 @@ MESSAGE(STATUS "Looking for Alembic. 1.5.8")
 ################################################################################
 # IlmBase include dir for half and ilm libraries used in alembic
 ################################################################################
-# Alembic includes half.h for a single function "half to float", this is sad
+
+# Alembic includes half.h for a single function "half to float", this is unfortunate
 FIND_PATH(ABC_HALF_INCLUDE_DIR half.h
     HINTS
     ${ALEMBIC_ILMBASE_ROOT}/include/OpenEXR
@@ -75,8 +76,7 @@ FIND_PATH(ABC_INCLUDE_DIR Alembic/Abc/All.h
 )
 
 #
-# We force the use of dynamic libraries as using the static ones had caused some 
-# initialization problems.
+# We force the use of dynamic libraries as with using the static ones we had initialization problems.
 FIND_PATH(ABC_LIBRARY_DIR libAlembicAbc.so
     PATHS
         ${ALEMBIC_ROOT}/lib
@@ -84,13 +84,6 @@ FIND_PATH(ABC_LIBRARY_DIR libAlembicAbc.so
         $ENV{ALEMBIC_ROOT}/lib
         $ENV{ALEMBIC_ROOT}/lib64
     NO_DEFAULT_PATH)
-#FIND_LIBRARY(ABC_COLLECTION libAlembicAbcCollection.so PATHS ${ABC_LIBRARY_DIR})
-#FIND_LIBRARY(ABC_COREFACTORY libAlembicAbcCoreFactory.so PATHS ${ABC_LIBRARY_DIR})
-#FIND_LIBRARY(ABC_COREOGAWA libAlembicAbcCoreOgawa.so PATHS ${ABC_LIBRARY_DIR})
-#FIND_LIBRARY(ABC_MATERIAL libAlembicAbcMaterial.so PATHS ${ABC_LIBRARY_DIR})
-#FIND_LIBRARY(ABC_OGAWA libAlembicOgawa.so PATHS ${ABC_LIBRARY_DIR})
-#FIND_LIBRARY(ABC_OPENGL libAlembicAbcOpenGL.so PATHS ${ABC_LIBRARY_DIR})
-#FIND_LIBRARY(ABC_WFOBJCONVERT libAbcWFObjConvert.so PATHS ${ABC_LIBRARY_DIR})
 FIND_LIBRARY(ABC libAlembicAbc.so PATHS ${ABC_LIBRARY_DIR})
 FIND_LIBRARY(ABC_COREABSTRACT libAlembicAbcCoreAbstract.so PATHS ${ABC_LIBRARY_DIR})
 FIND_LIBRARY(ABC_COREHDF5 libAlembicAbcCoreHDF5.so PATHS ${ABC_LIBRARY_DIR})
--- a/src/cmakeFindModules/FindAlembic.cmake
+++ b/src/cmakeFindModules/FindAlembic.cmake
@@ -3,9 +3,9 @@
 # You can help the search by providing several environment variable or cmake 
 # variable:
 # ALEMBIC_ROOT
-# ALEMBIC_HDF5_ROOT
-# ALEMBIC_ILMBASE_ROOT
-# ALEMBIC_OPENEXR_ROOT
+# HDF5_ROOT
+# ILMBASE_ROOT
+# OPENEXR_ROOT
 #
 # HDF5 and ILMBASE should point to the root dir used to compile alembic
 #
@@ -22,15 +22,15 @@ MESSAGE(STATUS "Looking for Alembic. 1.5.8")
 # Alembic includes half.h for a single function "half to float", this is unfortunate
 FIND_PATH(ABC_HALF_INCLUDE_DIR half.h
     HINTS
-    ${ALEMBIC_ILMBASE_ROOT}/include/OpenEXR
-    $ENV{ALEMBIC_ILMBASE_ROOT}/include/OpenEXR)
+    ${ILMBASE_ROOT}/include/OpenEXR
+    $ENV{ILMBASE_ROOT}/include/OpenEXR)
 
 FIND_PATH(ABC_ILMBASE_LIBS_PATH NAMES libIex.so libIex.a 
     PATHS
-        ${ALEMBIC_ILMBASE_ROOT}/lib 
-        ${ALEMBIC_ILMBASE_ROOT}/lib64
-        $ENV{ALEMBIC_ILMBASE_ROOT}/lib 
-        $ENV{ALEMBIC_ILMBASE_ROOT}/lib64
+        ${ILMBASE_ROOT}/lib 
+        ${ILMBASE_ROOT}/lib64
+        $ENV{ILMBASE_ROOT}/lib 
+        $ENV{ILMBASE_ROOT}/lib64
     NO_DEFAULT_PATH)
 
 FIND_LIBRARY(ABC_ILMBASE_IEX Iex PATHS ${ABC_ILMBASE_LIBS_PATH} NO_DEFAULT_PATH)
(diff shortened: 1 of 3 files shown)


src/software/SfM/main_openMVG2ABC.cpp
```diff
@@ -0,0 +1,168 @@
+#include <cstdlib>
```
pmoulon: Please include the MPL2 licence header and your copyright
```
// Copyright (c) 2015 cpichard.

// This Source Code Form is subject to the terms of the Mozilla Public
// License, v. 2.0. If a copy of the MPL was not distributed with this
// file, You can obtain one at http://mozilla.org/MPL/2.0/.
```

src/software/SfM/main_openMVG2ABC.cpp
```diff
@@ -0,0 +1,168 @@
...
+    //
+    SfM_Data sfm_data;
+    if (!Load(sfm_data, sfmDataFilename, ESfM_Data(ALL))) {
+        cout << "Error: The input project file \""<< sfmDataFilename << "\" cannot be read." << std::endl;
+        return EXIT_FAILURE;
+    }
+  
+    // Open Alembic archive
+    // TODO: decide wether we want to keep HDF5 or switch to Ogawa 
+    OArchive archive(Alembic::AbcCoreHDF5::WriteArchive(), sOutAlembic);
```
pmoulon: This could later be done using a command line option that allow to choose between HDF5 or Ogawa

src/software/SfM/main_openMVG2ABC.cpp
```diff
@@ -0,0 +1,168 @@
...
+        const openMVG::Vec3 &pt = lm.second.X;
+        positions.emplace_back(pt[0], pt[1], pt[2]);
+    }
+
+    std::vector<Alembic::Util::uint64_t> ids(positions.size());
+    iota(begin(ids), end(ids), 0);
+
+    OPoints partsOut(topObj, "particleShape1");
+    OPointsSchema &pSchema = partsOut.getSchema();
+    OPointsSchema::Sample psamp( move(V3fArraySample (positions)), move(UInt64ArraySample ( ids )));
```
pmoulon: std::move

src/software/SfM/main_openMVG2ABC.cpp
```diff
@@ -0,0 +1,168 @@
...
+          continue;
+
+        // OpenMVG Camera
+        const openMVG::geometry::Pose3 pose = sfm_data.GetPoseOrDie(view);
+        auto iterIntrinsic = sfm_data.GetIntrinsics().find(view->id_intrinsic);
+        openMVG::cameras::Pinhole_Intrinsic *cam = static_cast<openMVG::cameras::Pinhole_Intrinsic*>(iterIntrinsic->second.get());
+        openMVG::Mat34 P = cam->get_projective_equivalent(pose);
+
+        // Extract internal matrix, rotation and translation
+        openMVG::Mat3 R, K;
```
pmoulon: I recommend to use the rotation and the center from the pose:
```
const openMVG::Mat3 R = pose.R();
const openMVG::Vec3 center = pose.center();
```

src/software/SfM/main_openMVG2ABC.cpp
```diff
@@ -0,0 +1,168 @@
...
+        // Camera intrinsic parameters
+        OCamera camObj(xform, "camera_" + ss.str());
+        CameraSample camSample;
+    
+        // Take the max of the image size to handle the case where the image is in portrait mode 
+        const float imgWidth = cam->w();  
+        const float imgHeight = cam->h(); 
+        const float sensorWidthPix = std::max(imgWidth, imgHeight);
+        const float sensorHeightPix = std::min(imgWidth, imgHeight);
+        const float focalLengthPix = K(0,0);
```
pmoulon:
```
cam->focal()
```

src/software/SfM/main_openMVG2ABC.cpp
```diff
@@ -0,0 +1,168 @@
...
+        const float dx = K(0,2);
```
pmoulon:
```
cam->principal_point()(0);
```

src/software/SfM/main_openMVG2ABC.cpp
```diff
@@ -0,0 +1,168 @@
...
+        const float dy = K(1,2);
```
pmoulon:
```
cam->principal_point()(1);
```

src/software/SfM/main_openMVG2ABC.cpp
```diff
@@ -0,0 +1,168 @@
...
+        // OpenMVG Camera
+        const openMVG::geometry::Pose3 pose = sfm_data.GetPoseOrDie(view);
+        auto iterIntrinsic = sfm_data.GetIntrinsics().find(view->id_intrinsic);
+        openMVG::cameras::Pinhole_Intrinsic *cam = static_cast<openMVG::cameras::Pinhole_Intrinsic*>(iterIntrinsic->second.get());
+        openMVG::Mat34 P = cam->get_projective_equivalent(pose);
+
+        // Extract internal matrix, rotation and translation
+        openMVG::Mat3 R, K;
+        openMVG::Vec3 t;
+        openMVG::KRt_From_P(P, &K, &R, &t);
```
pmoulon: Do not use the decomposition, since we have the true value already stored

src/software/SfM/main_openMVG2ABC.cpp
```diff
@@ -0,0 +1,168 @@
...
+        xformMatrix[3][2]= center(2);
+        xformMatrix[3][3]= 1.0;
+
+        // Get correct camera orientation
+        M44d scale;
+        scale[0][0] = -1;
+        scale[1][1] = -1;
+        scale[2][2] = -1;
+
+        M44d scale2;
```
pmoulon: I think it assumes that by default the matrix is initialized with 0 values.

src/software/SfM/main_openMVG2ABC.cpp
```diff
@@ -0,0 +1,168 @@
...
+        << "[-o|--outfile path]\n"
+        << std::endl;
+        std::cerr << s << std::endl;
+        return EXIT_FAILURE;
+    }
+
+    //
+    SfM_Data sfm_data;
+    if (!Load(sfm_data, sfmDataFilename, ESfM_Data(ALL))) {
+        cout << "Error: The input project file \""<< sfmDataFilename << "\" cannot be read." << std::endl;
```
pmoulon: std::cout

src/software/SfM/main_openMVG2ABC.cpp
cpichard: Yes, this is definitely a good idea and a minor change. For the record, I am not quite sure if all the softwares used in post production can already read ogawa files, I still have to double check. So I guess it would be better to export hdf5 files by default for now.

src/software/SfM/main_openMVG2ABC.cpp
```diff
@@ -0,0 +1,168 @@
...
+        scale2[1][1] = 1;
+        scale2[2][2] = 1;
+        
+        xformMatrix = scale2*xformMatrix*scale;
+
+        XformSample xformsample;
+        xformsample.setMatrix(xformMatrix);
+
+        stringstream ss;
+        ss << stlplus::basename_part(view->s_Img_path);
```
pmoulon: Do you can link the camera to an image plane that will load the image corresponding to the view?

src/software/SfM/main_openMVG2ABC.cpp
cpichard: Indeed, the matrix is initialized with 1 in the diagonal and 0 everywhere else. See:
https://github.com/openexr/openexr/blob/master/IlmBase/Imath/ImathMatrix.h#L1904

src/software/SfM/main_openMVG2ABC.cpp
cpichard: I tested with "pose.R()" but it is not in implemented in the version that I am using. Is this new function on the current develop branch ?

pmoulon: Oups sorry, it is pose.rotation() :+1:

src/software/SfM/main_openMVG2ABC.cpp
cpichard: I suppose we can add an extra field in the camera object to keep the image file name but I still have to investigate if it is really possible. It might be a good addition though.

src/software/SfM/main_openMVG2ABC.cpp
cpichard: thx !

src/software/SfM/CMakeLists.txt
```diff
@@ -128,6 +128,15 @@ TARGET_LINK_LIBRARIES(openMVG_main_openMVG2PMVS ${OpenMVG_LIBRARIES})
 ADD_EXECUTABLE(openMVG_main_openMVG2CMPMVS main_openMVG2CMPMVS.cpp)
 TARGET_LINK_LIBRARIES(openMVG_main_openMVG2CMPMVS ${OpenMVG_LIBRARIES})
 
+# - Export a SfM openMVG scene to Alembic format
+#
+FIND_PACKAGE(Alembic)
+if (ALEMBIC_FOUND)
+  ADD_EXECUTABLE(openMVG_main_openMVG2ABC main_openMVG2ABC.cpp)
+  INCLUDE_DIRECTORIES(${ABC_INCLUDE_DIR})
+  TARGET_LINK_LIBRARIES(openMVG_main_openMVG2ABC ${ABC_LIBRARIES} ${OpenMVG_LIBRARIES})
```
fabiencastan: There is a missing INSTALL rule.

pmoulon:
I project take a look to it soon.
I have try to setup only alembic...locally... but http://docs.alembic.io/python/abcview.html configure seems not work with local and only consider system installed libraries...

src/cmakeFindModules/FindAlembic.cmake
```diff
@@ -0,0 +1,107 @@
...
+#
+# We force the use of dynamic libraries as with using the static ones we had initialization problems.
+FIND_PATH(ABC_LIBRARY_DIR libAlembicAbc.so
+    PATHS
+        ${ALEMBIC_ROOT}/lib
+        ${ALEMBIC_ROOT}/lib64
+        $ENV{ALEMBIC_ROOT}/lib
+        $ENV{ALEMBIC_ROOT}/lib64
+    NO_DEFAULT_PATH)
+FIND_LIBRARY(ABC libAlembicAbc.so PATHS ${ABC_LIBRARY_DIR})
```
simogasp: @cpichard I think the standard way to specify the lib name for FIND_LIBRARY is without the prefix `lib` and the extension `.so` (for Mac, eg would be .dylib)

src/software/SfM/CMakeLists.txt
cpichard: Added in the last commit !

PR #371, opened 2015-08-21
Multiple LEMON library naming hints.

cbalint13:
Folks,

 One more tiny fix, one single line. LEMON on fedora system is:
-----------<<<<---------------
$ rpm -ql coin-or-lemon | grep so
/usr/lib64/libemon.so.1.3.1
----------->>>>---------------
* Notice that lib name is "emon".
* On MSWIN i understand it could be Lemon, or lemon.


pmoulon:
Thanks, merged by https://github.com/openMVG/openMVG/commit/9ac8872658878a38b40c3ecb18154f5bcdbfd81e
emon is less sexy that lemon ;-)

PR #374, opened 2015-08-24
minor fix in html report exporation.

baritone:
More secure way to create the HTML report when using non central cameras. Indeed, some views belongs to a pose, but they don't see any 3D point.


pmoulon:
Thanks. Closed by https://github.com/openMVG/openMVG/commit/ab8bc10a8d851015e73c5774d051e1516c159dfd

PR #376, opened 2015-08-24
Option to use UID instead of index for cameras

fabiencastan:
Will be useful to merge reconstructions, add images, etc.


src/openMVG/exif/exif_IO_EasyExif.hpp
```diff
@@ -94,9 +94,48 @@ class Exif_IO_EasyExif : public Exif_IO
...
+    // Original file date and time (may not exist)
+    std::string getDateTimeOriginal() const
+    {
+      return exifInfo_.SubSecTimeOriginal;
+    }
+
+    // Digitization date and time (may not exist)
+    std::string getDateTimeDigitized() const
+    {
+      return exifInfo_.SubSecTimeOriginal;
```
pmoulon: copy/paste ;-)
```
exifInfo_.DateTimeDigitized
```

src/openMVG/exif/exif_IO_EasyExif.hpp
```diff
@@ -94,9 +94,48 @@ class Exif_IO_EasyExif : public Exif_IO
...
+    // File change date and time
+    std::string getDateTime() const
+    {
+      return exifInfo_.DateTime;
+    }
+
+    // Original file date and time (may not exist)
+    std::string getDateTimeOriginal() const
+    {
+      return exifInfo_.SubSecTimeOriginal;
```
pmoulon: copy/paste ;-)
```
exifInfo_.DateTimeOriginal
```

src/openMVG/exif/exif_IO_EasyExif.hpp
```diff
@@ -110,41 +149,45 @@ class Exif_IO_EasyExif : public Exif_IO
...
+        << "Image width        : " << exifInfo_.ImageWidth << "\n"
+        << "Image height       : " << exifInfo_.ImageHeight << "\n"
+        << "Image description  : " << exifInfo_.ImageDescription << "\n"
+        << "Image orientation  : " << exifInfo_.Orientation << "\n"
+        << "Image copyright    : " << exifInfo_.Copyright << "\n"
+        << "Image date/time    : " << exifInfo_.DateTime << "\n"
+        << "Original date/time : " << exifInfo_.DateTimeOriginal << "\n"
+        << "Digitize date/time : " << exifInfo_.DateTimeDigitized << "\n"
+        << "Subsecond time     : " << exifInfo_.SubSecTimeOriginal << "\n"
+        << "Exposure time      : 1/time " << (unsigned) (1.0/exifInfo_.ExposureTime) << "\n"
```
pmoulon: does it will appear as a float... we must see "1/" << 4 ?

src/software/SfM/main_SfMInit_ImageListing.cpp
```diff
@@ -98,6 +100,7 @@ int main(int argc, char **argv)
       << "[-g|--group_camera_model]\n"
       << "\t 0-> each view have it's own camera intrinsic parameters,\n"
       << "\t 1-> (default) view can share some camera intrinsic parameters\n"
+      << "[-u|--use_UID] Generate a UID instead of using indexes\n"
```
pmoulon: Perhaps the help message must be more clear (we must mention view):
Something in the vein of: View indices : UID (unique id) or incremental numbers.

src/software/SfM/main_SfMInit_ImageListing.cpp
```diff
@@ -271,8 +275,15 @@ int main(int argc, char **argv)
       }
     }
 
+    IndexT id_view = views.size();
+    if( b_use_UID )
+    {
+      std::size_t uid = computeUID(exifReader, imageFilename);
```
pmoulon: const std::si..

src/third_party/easyexif/exif.h
```diff
@@ -95,6 +96,10 @@ class EXIFInfo {
   std::string DateTimeDigitized;    // Digitization date and time (may not exist)
   std::string SubSecTimeOriginal;   // Sub-second time that original picture was taken
   std::string Copyright;            // File copyright information
+  std::string ImageUniqueID;
```
pmoulon: Unique test for those new exif tag would be great...

fabiencastan:
The PR is better now.
It's not perfect because easyexif doesn't deal with MNOTE tags which are vendor specific.
So the PR is fine because if the generic metadatas is not there (as it's the case in the sample image used in unittests), it fallbacks on the name and date.

In the future, we may need to switch to [libexif](http://libexif.sourceforge.net/).

PR #402, opened 2015-09-30
Avoid creating empty view directories

rhiestan:
dmrecon chokes on empty view directories.

The code compiles, however I was not able to test main_openMVG2MVE2.exe because the current develop branch does not compile on my PC.
However, I tested the code in my project and it works fine with dmrecon.


pmoulon:
Thank you. :+1:

PR #415, opened 2015-10-11
fix issue #413 for zlib

### LINKED ISSUE #413: zlib not included when PNG is external and TIFF is internal
A linking problem arises when an external PNG lib is detected while TIFF is compile internally.
In thirdparty/CMakeLists we need to deal with this case.
https://github.com/openMVG/openMVG/blob/develop/src/third_party/CMakeLists.txt#L58

simogasp:
TIFF and PNG depend on zlib, if one of them is not found add the
internal zlib. This fixes a linking issue when PNG is external and TIFF is internal, in that case zlib was not included as dependency to build. 
fixes #413


pmoulon:
Do we must close #413?

pmoulon:
Done with commit https://github.com/openMVG/openMVG/commit/62dfac227a56a3c5e39abdb41a7dd2266bc53184

PR #427, opened 2015-11-04
Reduce point feature memory footprint

cpichard:
Hi Pierre,

here is a simple change to reduce the size of a point feature in memory. It basically removes the memory associated with the virtual table of the PointFeature and SIOPointFeature classes. Even if it's not a lot (8 bytes per feature) I noticed a small time performance gain of approximately 1% on feature extraction and matching. I am going to test further the repercussions of this change. Let me know if this looks good for you !

Cheers,

Cyril


pmoulon:
Your change looks ok.
Are you able to provide more feedback as you have mentioned it...

PR #444, opened 2015-12-29
parallel features extraction using subprocesses on linux

cpichard:
Hi Pierre,

This pull request allows to run multiple features extraction sub-processes in main_ComputeFeatures using -j command line option. This method is faster on our machines, up to x6 depending on the number of jobs, but with the increase of speed comes the increase of memory consumptionm so there is a tradeof.
Please let me know if that might interest you, I'll be happy if you could test this PR on your system to check if this gives better feature extraction speed.   

Best,

Cyril

### DIFF
--- a/src/software/SfM/main_ComputeFeatures.cpp
+++ b/src/software/SfM/main_ComputeFeatures.cpp
@@ -55,56 +55,8 @@ features::EDESCRIBER_PRESET stringToEnum(const std::string & sPreset)
 #include <unistd.h>
 #include <sys/wait.h>
 
-// This function dispatch the compute function on several sub processes
-// keeping the maximum number of subprocesses under maxJobs
-void dispatch(const int &maxJobs, std::function<void()> compute)
-{
-  static int nbJobs;
-  pid_t pid = fork();
-  if (pid < 0)           
-  {                      
-    // Something bad happened
-    std::cerr << "fork failed\n";
-    _exit(EXIT_FAILURE);
-  }
-  else if(pid == 0)
-  {
-#ifdef OPENMVG_USE_OPENMP
-    // Disable OpenMP as we dispatch the work on multiple sub processes
-    // and we don't want that each subprocess use all the cpu ressource
-    omp_set_num_threads(1); 
-#endif
-    compute();
-    _exit(EXIT_SUCCESS);
-  }
-  else 
-  {
-    nbJobs++;
-    if (nbJobs && (nbJobs % maxJobs == 0))
-    {
-      pid_t pids;
-      while(pids = waitpid(-1, NULL, 0)) 
-      {
-        nbJobs--;
-        break;
-      }
-    }
-  }
-}
-
-// Waits for all subprocesses to be completed
-void waitForCompletion()
-{
-  pid_t pids;
-  while(pids = waitpid(-1, NULL, 0)) 
-  {
-    if (errno == ECHILD)
-        break;
-  }
-}
-
-// Returns a map containing information about the memory usage,
-// it basically reads /proc/meminfo  
+// Returns a map containing information about the memory usage o
+// of the system, it basically reads /proc/meminfo  
 std::map<std::string, unsigned long> memInfos()
 {
   std::map<std::string, unsigned long> memoryInfos;
@@ -126,7 +78,8 @@ std::map<std::string, unsigned long> memInfos()
   return memoryInfos;
 }
 
-// Count the number of processors on the machine using /proc/cpuinfo
+
+// Count the number of processors of the machine using /proc/cpuinfo
 unsigned int countProcessors()
 {
   unsigned int nprocessors = 0;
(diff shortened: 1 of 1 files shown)


src/software/SfM/main_ComputeFeatures.cpp
```diff
@@ -65,31 +131,36 @@ int main(int argc, char **argv)
...
+    << "   SIFT (default),\n"
+    << "   AKAZE_FLOAT: AKAZE with floating point descriptors,\n"
+    << "   AKAZE_MLDB:  AKAZE with binary descriptors\n"
+    << "[-u|--upright] Use Upright feature 0 or 1\n"
+    << "[-p|--describerPreset]\n"
+    << "  (used to control the Image_describer configuration):\n"
+    << "   NORMAL (default),\n"
+    << "   HIGH,\n"
+    << "   ULTRA: !!Can take long time!!\n"
+    << "[-j|--jobs] Specifies the number of jobs to run simultaneously\n"
```
pmoulon: Does this option must appear only on linux like system?

cpichard: Yes indeed, it is not relevant on mac and windows yet. 
I'll fix that next week, when I am back in Paris.
Thanks

cpichard:
Hi,

with the latest commit, computeFeatures should now work as before, on linux/win/mac. 
When the -j N command line option is used, computeFeatures launches N subprocesses on linux. On other operating systems, -j N is used to control the number of openmp threads, provided computeFeatures was compiled with openMVG support.

Cheers,

cpichard:
Hi Pierre,
I am not sure I get the point. I used constexpr as I wanted this variable to be evaluated at compile time, but if you prefer const I have no problem changing it.

Best

C

fabiencastan:
@cpichard Why are you using subprocess instead of multi-threading with openmp? If you use openmp on a top level for loop (on the list of images), neested usages of openmp are disabled by default.

cpichard:
Hi @fabiencastan ,

In the case of feature extraction, it is way faster to parallelize using sub-processes instead of using openmp. The tests done on several machine gives speedup up to x5, which is not negligible. Each feature extraction process spend roughly 20% of its time in the omp parallel section, the rest is spent in reading/writing/converting and so on. Using this method gives a straightforward way of interleaving reading/writing with computation and have N feature extraction working at the same time. The only downside is memory requirement, multiplied by the number of subprocesses N. For a 5K image, the memory required is around 2G.

C

pmoulon:
Have you tried simply use GNU parallel?

fabiencastan:
Yes, I understand that it is more efficient to parallelize on the high level instead of relying on vlfeat low level openmp usage. But we can parallelize to compute multiple images in parallel using threads instead of processes. What is the reason to use processes instead of threads?

cpichard:
Hi,
nothing forbids to use threads instead of processes, I chose the sub processes on linux for the following reasons:
  - It was an easy and fast way to implement a safe dispatcher,
    - we don't need to share memory between several features extraction, so each computation can be totally independent.
    - by being independent, we can easily use a process to encapsulate/protect the memory: when one subprocess crash, it does not crash the whole application 
    - no possible memory corruption as each subprocess has its own virtual memory
  - A subprocess having problems can be easily monitored/killed by any sysadmin or developer
  - It potentially works with and without openmp
  - Given the granularity of the feature extraction, it seemed to me that a subprocess was more appropriate than openmp threads.                              

We could try to dispatch with C++ 11 threads, it might be a solution to dispatch on windows, but I have no idea how it is going to interact with openmp yet.

C

autosquid:
One more point to supplement: sub-processes enable feature extraction to grab more computing resources.

pmoulon:
Since there is already some multithreading in the SIFT descriptor extraction.
Perhaps we must consider put the thread count to 1 when using sub-processes else some multithread task will collide.

cpichard:
Hi Pierre,

yes indeed, there is a omp_set_num_threads(1) just before the compute call that does exactly that, otherwise the load on the machine could be too high and we potentially loose efficiency.

pmoulon:
Thanks, I did'nt notice it ;-) :+1:

fabiencastan:
Does it make sense to detect the number of processors on the machine and divide by the number of subprocesses instead of setting it to 1?

cpichard:
Well, omp_set_num_threads takes an integer which means that it has to be at least 2 to make sense. In that configuration you would have half of the processors working on feature extraction and the other half helping feature extraction for only 10% of the time. It might be better but not up to an order of magnitude, we could test to have some numbers. However, in the general case, when the memory allows it, I think it's better to keep the number of processor close to the number of running processes.

cpichard:
Hi Pierre,

I added the automatic number of job selection based on the premise that 1 job should run on 1 proc and 2GB of memory. This mode is selected on the command line with the option -j 0. Let me know if this option is of interest to you.

Best,

Cyril

cpichard:
Hi Pierre,
As you are developing your own version of this feature I suppose that this pull request will never be merged. So can I close it ? 
Best,
C

pmoulon:
Hi @cpichard 
It does not mean I don't want to merge it.

I would prefer first a simple solution then create a more complex binary that could have better performance (like you have done).

The solution that you ended too is a bit different than classic threading (openmp) and so it make the code less readable (at first sight).
So I think it could be considered as a good feature for v1.1

What do you think about my point?

cpichard:
Hi Pierre,

up to you, I keep the pull request open then. But no worries, it's obvious that the openmp solution is more elegant, easy to understand and moreover as fast as the subprocess version. I was just asking this because I wanted to do some "github" cleaning.
This subprocess version is really worth if you want to include external feature extractor without spending time trying to find threading and memory issues or if you need, as we do, to run feature extractions on an heterogeneous park of machine and automatically detect the best number of threads.

Cheers,

C

Issue #452, opened 2016-01-19
Occasional incorrect intial pair selection


donlk:
Ok, i just bumped into this.

In sequential_SfM.cpp if automatic initial pair selection fails, the software calls for manual intervention. After a selection has been made, the reference parameter contains the desired initial pair, but it does not get passed for robust initial computation, as it happens for the automatically chosen pair. The local pair variable does not get overriden, resulting in occasional Segmentation Faults as tracks get selected between image indexes (0, 0), and the software assumes 2 unique index pairs in  MakeInitialPair3D(...).

Problematic section:
```
// Initial pair choice
  Pair initialPairIndex = _initialpair;
  if (_initialpair == Pair(0,0))
  {
    Pair putative_initial_pair;
    if (AutomaticInitialPairChoice(putative_initial_pair))
    {
      initialPairIndex = _initialpair = putative_initial_pair;
    }
    else // Cannot find a valid initial pair, try to set it by hand?
    {
      if (!ChooseInitialPair(_initialpair))
        return false;
    }
  }
  // Else a starting pair was already initialized before

  // Initial pair Essential Matrix and [R|t] estimation.
  if (!MakeInitialPair3D(initialPairIndex))
    return false;
```

Simple fix:
```
if (ChooseInitialPair(_initialpair)){
	initialPairIndex = _initialpair;
}else{
	return false;
}
```

donlk:
Would you like a pull request?

pmoulon:
Thanks for having found this bug.
You can perform a pull request on the develop branch.

The other way to see the problem can be to remove the initialPairIndex variable and only consider the _initialpair one.
So remove line https://github.com/openMVG/openMVG/blob/master/src/openMVG/sfm/pipelines/sequential/sequential_SfM.cpp#L96 and replace initialPairIndex by _initialpair on line https://github.com/openMVG/openMVG/blob/master/src/openMVG/sfm/pipelines/sequential/sequential_SfM.cpp#L113

pmoulon:
Closed by merging #453

PR #453, opened 2016-01-19
Fix for occasionally incorrect intial pair selection

donlk:
Hi pierre,
Here's the pull request for #452. I took the liberty of using only one type of initial pair variable to make it less confusing.


pmoulon:
Merged. Thanks @donlk :+1:

PR #474, opened 2016-02-16
Develop parameters

hcjghr:
Hi!

This pull request adds the possibility of setting the parameters (for initialization, feature detection, matching and global/incremental SfM) in a JSON file. This facilitate saving the settings used in reconstructions and easier re-usage. For now only the parameters that could be set in the command line are included but I want to also add some of the parameters currently hard-coded (thresholds in SfM) to this set in order to make it easier to change the values depending on the dataset but I think it is better to do it in two steps.

Looking forward to your remarks.

Klemen


src/openMVG/params/params_Detection.hpp
```diff
@@ -0,0 +1,59 @@
...
+	//		- SIFTGPU
+	// 		- AKAZE_FLOAT: AKAZE with floating point descriptors
+	//		- AKAZE_MLDB:  AKAZE with binary descriptors
+	std::string feature_type;
+
+	// Feature detector configuration (simple way)
+	//		- NORMAL (default)
+	//		- HIGH
+	//		- ULTRA !!Can take long time!!
+	std::string feature_present;
```
simogasp: `feature_preset`

src/openMVG/params/params_Matching.hpp
```diff
@@ -0,0 +1,85 @@
...
+	    {}
+
+	  // Serialization
+	  template <class Archive>
+	  void serialize( Archive & ar )
+	  {
+	    ar(cereal::make_nvp("max_matching_dist_ratio", max_matching_dist_ratio),
+	    	cereal::make_nvp("geometric_model", geometric_model),
+	    	cereal::make_nvp("video_mode_matching", video_mode_matching),
+			cereal::make_nvp("nearest_matching_method", nearest_matching_method),
```
pmoulon: tab alignement

src/software/SfM/main_ComputeMatches.cpp
```diff
@@ -127,6 +134,32 @@ int main(int argc, char **argv)
...
+  // Try to open file if path is provided
+  if (!sParams_Data_Filename.empty()){
+	  if (!Load(params_data, sParams_Data_Filename)) {
+		std::cout << std::endl
+		  << "The input parameters file \""<< sParams_Data_Filename << "\" cannot be read." << std::endl;
+	  }
+	  else{
+		  // Set loaded parameters
+		  std::cout << std::endl
+		  		  << "Parameters loaded from: \""<< sParams_Data_Filename << "\"" << std::endl << std::endl;
```
pmoulon: tab alignement

pmoulon:
Hi @hcjghr 
Since parameters configuration are only used in software, I think that the params module should be moved from src/openMVG/params to src/software/SfM/params.

The other question that remains is how parameters override the other.
I mean, I use some from the JSON and use some from the command line... What is the expected behavior in this case?

simogasp:
> Since parameters configuration are only used in software, I think that the params module should be moved from src/openMVG/params to src/software/SfM/params.
Or, for example, `SequentialSfMReconstructionEngine` could (optionally) take `paramsIncrementalSfM` as input to set all the parameters that will be later exposed.

hcjghr:
@pmoulon my idea was that if you specify the parameter file in the command line, the parameter values set in it have priority over the values set directly in the command line. Of course this is just my suggestion and it could be the other way around.

hcjghr:
I have added more parameters for matching, incremental and global SfM. @pmoulon currently I left the params in src/openMVG/param as now I also use it in the _SequentialSfMReconstructionEngine_ and  _GlobalSfMReconstructionEngine_RelativeMotions_ in order to make the use of the parameters for custom threshold values for SfM process. Do you agree with this?

PR #478, opened 2016-02-23
Adding doxygen documentation to openMVG

rperrot:
I added doxygen documentation generation to openMVG and some comments on various classes. I think it could be useful for new developers.


pmoulon:
Just for info, I prefer PR from develop to develop.
Since master is only used for finale release.

PR #490, opened 2016-03-03
parallel feature extraction

mitjap:
Adds an option to run feature extraction in parallel.

### DIFF
--- a/src/software/SfM/main_ComputeFeatures.cpp
+++ b/src/software/SfM/main_ComputeFeatures.cpp
@@ -22,6 +22,10 @@
 #include <cstdlib>
 #include <fstream>
 
+#ifdef OPENMVG_USE_OPENMP
+#include <omp.h>
+#endif
+
 using namespace openMVG;
 using namespace openMVG::image;
 using namespace openMVG::features;
@@ -56,6 +60,9 @@ int main(int argc, char **argv)
   std::string sImage_Describer_Method = "SIFT";
   bool bForce = false;
   std::string sFeaturePreset = "";
+#ifdef OPENMVG_USE_OPENMP
+  int iNumThreads = 0;
+#endif
 
   // required
   cmd.add( make_option('i', sSfM_Data_Filename, "input_file") );
@@ -65,6 +72,10 @@ int main(int argc, char **argv)
   cmd.add( make_option('u', bUpRight, "upright") );
   cmd.add( make_option('f', bForce, "force") );
   cmd.add( make_option('p', sFeaturePreset, "describerPreset") );
+
+#ifdef OPENMVG_USE_OPENMP
+  cmd.add( make_option('n', iNumThreads, "numThreads") );
+#endif
 
   try {
       if (argc == 1) throw std::string("Invalid command line parameter.");
@@ -86,6 +97,9 @@ int main(int argc, char **argv)
       << "   NORMAL (default),\n"
       << "   HIGH,\n"
       << "   ULTRA: !!Can take long time!!\n"
+#ifdef OPENMVG_USE_OPENMP
+      << "[-n|--numThreads] number of parallel computations\n"
+#endif
       << std::endl;
 
       std::cerr << s << std::endl;
@@ -99,7 +113,11 @@ int main(int argc, char **argv)
             << "--describerMethod " << sImage_Describer_Method << std::endl
             << "--upright " << bUpRight << std::endl
             << "--describerPreset " << (sFeaturePreset.empty() ? "NORMAL" : sFeaturePreset) << std::endl
-            << "--force " << bForce << std::endl;
+            << "--force " << bForce << std::endl
+#ifdef OPENMVG_USE_OPENMP
+            << "--numThreads " << iNumThreads << std::endl
+#endif
+            << std::endl;
 
 
   if (sOutDir.empty())  {
@@ -213,11 +231,14 @@ int main(int argc, char **argv)
     Image<unsigned char> imageGray;
     C_Progress_display my_progress_bar( sfm_data.GetViews().size(),
       std::cout, "\n- EXTRACT FEATURES -\n" );
-    for(Views::const_iterator iterViews = sfm_data.views.begin();
-        iterViews != sfm_data.views.end();
-        ++iterViews, ++my_progress_bar)
+
+#ifdef OPENMVG_USE_OPENMP
+    omp_set_num_threads(iNumThreads);
+    #pragma omp parallel for schedule(dynamic) firstprivate(imageGray) if(iNumThreads > 0)
+#endif
+    for(int i = 0; i < sfm_data.views.size(); ++i)
     {
-      const View * view = iterViews->second.get();
+      const View * view = sfm_data.views[i].get();
       const std::string sView_filename = stlplus::create_filespec(sfm_data.s_root_path,
         view->s_Img_path);
       const std::string sFeat = stlplus::create_filespec(sOutDir,
@@ -236,6 +257,12 @@ int main(int argc, char **argv)
         image_describer->Describe(imageGray, regions);
         image_describer->Save(regions.get(), sFeat, sDesc);
       }
+
+#ifdef OPENMVG_USE_OPENMP
+    #pragma omp critical
+#endif
+    ++my_progress_bar;
+
     }
     std::cout << "Task done in (s): " << timer.elapsed() << std::endl;
   }


pmoulon:
Have you checked that the output produced files looks exactly the same (md5sum)?
When I have tried in the past to do that I got different results between some run, since VLFEAT SIFT is not multi-thread safe.

mitjap:
Thank you for pointing that out. I will do md5 checks and report back. Will also take another look at VLFEAT's SIFT implementation.

mitjap:
Output (md5) is different on every run for both solutions, single and multi-threaded. After some cross-checking I noticed that rows are shuffled. I will create a "validator" that allows rows in files to be in different order. File size is consistent across multiple runs.

pmoulon:
Perhaps if you comment this you will have no more row shuffling https://github.com/openMVG/openMVG/blob/master/src/nonFree/sift/SIFT_describer.hpp#L156

But when I tried to parallel them the run was not stable. I think it's because VLFEAT is not multithread safe, since I initialize it many times here https://github.com/openMVG/openMVG/blob/master/src/nonFree/sift/SIFT_describer.hpp#L123

Perhaps using a singleton with a mutex on it can make it safe...

mitjap:
I ran the tests with disabled shuffling and checksums are identical. I tested multiple times with 1, 4, 6 and 8 threads.

mitjap:
It might be good idea to _not_ run [SIFT_describer.hpp#L156](https://github.com/openMVG/openMVG/blob/master/src/nonFree/sift/SIFT_describer.hpp#L156) `for loop` in parallel if already in parallel block. Check can be done with [omp_in_parallel()](https://computing.llnl.gov/tutorials/openMP/#OMP_IN_PARALLEL).

cpichard:
Hi @mitjap,

You might want to have a look at the the parallel features extraction version we use in our company:
https://github.com/openMVG/openMVG/pull/444
It is based on subprocesses and only works on linux for now. The code is definitely bigger than your clearer and concise OpenMP approach but we chose to use subprocesses instead of threads/openMP cause we couldn't be sure that the current and future feature extraction code is going to be thread safe.
Anyway, I am very curious to see how both approaches compares in terms of performance, I'll try today if I have time.

Cyril

pmoulon:
@cpichard Your performance test would be very interesting ;-)

cpichard:
Hi,
I tried the OpenMP version, it behaves well with akaze, it is as fast as the subprocess version, even a bit faster. Unfortunately I can't get the sift method to work. The computation starts correctly with the number of threads specified, and then after the first batch of images, the number of working thread decreases radically and continue to decrease to the point where there are no thread working at all and the process gets stuck. Did you have this kind of problem on your end ?

C

pmoulon:
@cpichard Did you comment this line https://github.com/openMVG/openMVG/blob/master/src/nonFree/sift/SIFT_describer.hpp#L156
In order to avoid to have an openMP thread in existing openMP threads...

cpichard:
@pmoulon: I just did and I got the same issue :(

fabiencastan:
Yes, nested loops are disabled by default in openmp:

`The OMP_NESTED environment variable enables or disables nested parallelism unless nested parallelism is enabled or disabled by calling the omp_set_nested library routine (see the omp_nested(3) man page). If set to TRUE, nested parallelism is enabled; if it is set to FALSE, nested parallelism is disabled. The default value is FALSE.
`

src/software/SfM/main_ComputeFeatures.cpp
```diff
@@ -213,11 +231,14 @@ int main(int argc, char **argv)
...
-        ++iterViews, ++my_progress_bar)
+
+#ifdef OPENMVG_USE_OPENMP
+    omp_set_num_threads(iNumThreads);
+    #pragma omp parallel for schedule(dynamic) firstprivate(imageGray) if(iNumThreads > 0)
+#endif
+    for(int i = 0; i < sfm_data.views.size(); ++i)
     {
-      const View * view = iterViews->second.get();
+      const View * view = sfm_data.views[i].get();
```
pmoulon: In case of none contiguous index in sfm_data it's better to use something like:
```
Views::const_iterator iterViews = sfm_data.views.begin();
std::advance(iterViews, i);
const View * view = iterViews->second.get();
```

src/software/SfM/main_ComputeFeatures.cpp
```diff
@@ -236,6 +257,12 @@ int main(int argc, char **argv)
         image_describer->Describe(imageGray, regions);
         image_describer->Save(regions.get(), sFeat, sDesc);
       }
+
```
pmoulon: Since there is a continue sooner in the loop, this progress bar can have a unvalid value if one image cannot be open.
Will be better if we replace line 252 too
```
        Image<unsigned char> imageGray;
        if (ReadImage(sView_filename.c_str(), &imageGray))
        {
          // Compute features and descriptors and export them to files
          std::unique_ptr<Regions> regions;
          image_describer->Describe(imageGray, regions);
          image_describer->Save(regions.get(), sFeat, sDesc);
        }
```

src/software/SfM/main_ComputeFeatures.cpp
```diff
@@ -56,6 +60,9 @@ int main(int argc, char **argv)
   std::string sImage_Describer_Method = "SIFT";
   bool bForce = false;
   std::string sFeaturePreset = "";
+#ifdef OPENMVG_USE_OPENMP
+  int iNumThreads = 0;
```
pmoulon: Having the default value equals to 1 would be preferable (default thread count should be a positive integer)?

mitjap: I was using `iNumThreads  = 0` to disable parallel block entirely on line 237.  [omp_in_parallel()](https://msdn.microsoft.com/en-us/library/y22bxh22.aspx) would return false in "parallelized" for loop, and would behave exactly the same as non-parallelized version. Also nested for loop with existing parallelization should run in parallel, which they would normally not if `omp_in_parallel` returns true/nonzero - except if explicitly set with [omp_set_nested(int)](https://msdn.microsoft.com/en-us/library/sk3zt8e1.aspx). But yeah, I did not check whether `omp_set_num_threads(int)` can receive 0 as input. Its probably best to set default to 1.

pmoulon:
I have pushed a new branch called develop_parallel_feature_extraction that contains @mitjap modification and mine (98e60d9ca5289bcd4c98ea59449bcf11c0a53429)
Behavior observed with SIFT by @cpichard must not longer appear (speed decreasing).

FYI @igorti @pyp22 @ORNis @rperrot
If you want use this new option, use -n X (X indicating the number of core you want to use, 2 - 4 - 6 - 8 ...)
=> note that if you use very large image and a large number of core you can have a segfault...

@mitjap you would be very kind to test that I have not broke anything with my additional commits...

rperrot:
I perfectly understand the need. But the solution is (in my oppinion), not a good idea. Multithreading should be done at extraction/description scale not at image scale. I know this should result either on modification of the vl_feat code, either adding a new implementation of SIFT extraction but (again in my oppinion) may result in a more scalable code. 

I'm not saying everybody did a mistake but I think this should not be a long term feature (it's good because it fix a lack and a need and I'm OK with that).

I did not run a test, and probably won't do it because using single thread SIFT with -p ULTRA uses  about 10Gb of RAM, with -n 2 this may results in more than the actual memory of my computer (only 16Gb), which (if it runs without segfault) may result in slower code than one single thread.

pmoulon:
Thanks for your opinion sharing @rperrot 

My idea is that:
-> if the Image_descriptor code use already some multithreading (if so, run with one thread will be ok)
-> if the Image_descriptor interface is not parallel and does not use much memory (run many thread is not a problem) -> like the TBMR/MSER/CORNER extraction interfaces.

So we must keep in mind that the two kind of parallelism can be useful.
But I agree there is not perfect solution for the moment.

cpichard:
Hi Pierre,

Sorry I didn't get back to you sooner but I used intel inspector to look for data races and other sorts of issues few weeks ago. It didn't work very well, inspector hung on every run, I spare you the details, but I came up with several changes that you can find in the attached patch. 
[feature_extraction_omp.diff.zip](https://github.com/openMVG/openMVG/files/197678/feature_extraction_omp.diff.zip)

Some are similar to your latest changes  [,](https://github.com/openMVG/openMVG/commit/98e60d9ca5289bcd4c98ea59449bcf11c0a53429) and there is also a fix for a global variable accessed by threads reading and writing to it in vlsift. You might be interested by this patch if you want to go for the omp solution. 

C

pmoulon:
Hi @cpichard 
Thank you for your in depth check.
From your change I don't understand why fast_expn_init call have been removed from vl_sift_new(...)
The LUT table still need to be initialized....

cpichard:
correct ! it has to be initialized !

pmoulon:
Branch is merged.
I just restore the old behavior (by default the Image_describer can use all the thread), but it is better to use the -n X.
Thank you @mitjap

igorti:
I've been using it on few datasets and besides high memory consumption(1-2GB per thread), which is to expect, it works very good. It's a great addition since it allows to run through big datasets in reasonable amount of time. Thanks!

mitjap:
I'm really glad to see this feature merged. And sorry for my inactivity regarding these changes lately.

pmoulon:
@mitjap No worries, you made the first step and then the community was here to react (testing & validating). Hoping that other PR will appear from you later on ;-) (I'm sure you saw that the GCP support is almost complete).
@igorti Appreciate your comment regarding the validation.

PR #512, opened 2016-03-27
interface: openMVG to openMVS

cdcseacave:
thx

### DIFF
--- a/src/software/SfM/CMakeLists.txt
+++ b/src/software/SfM/CMakeLists.txt
@@ -208,6 +208,17 @@ TARGET_LINK_LIBRARIES(openMVG_main_openMVG2PMVS
   stlplus
   )
 
+# - Export a SfM openMVG scene to openMVS format
+#
+ADD_EXECUTABLE(openMVG_main_openMVG2openMVS main_openMVG2openMVS.cpp)
+TARGET_LINK_LIBRARIES(openMVG_main_openMVG2openMVS
+  openMVG_system
+  openMVG_image
+  openMVG_features
+  openMVG_sfm
+  stlplus
+  )
+
 # - Export a SfM openMVG scene to CMPMVS format
 #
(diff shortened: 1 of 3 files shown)


pmoulon:
Thanks @cdcseacave for this PR.
I think that the openMVGtoOpenMVS exporter must be compiled with the same options in order to work.... I mean the _USE_BOOST & _USE_BOOST_SERIALIZATION option must be set to the same value?
Having a transparent handling of the two format would be a nice addition to OpenMVS. Perhaps we could detect the BOOST SERIALIZATION automatically since the file is Zipped... there is certainly a magic header we can check about...

See here the hard coded check https://github.com/cdcseacave/openMVS/blob/master/libs/MVS/Scene.cpp#L65

cdcseacave:
You are right, however the BOOST serialization is there only for historic reasons, it is not and should be never used. In fact I just forgot to delete it, before submitting the PR. I will submit a new PR.

pmoulon:
Thanks for your explaination.
You can update the code in your branch and squash your commits (in order to have only one) and repush to your branch using force, the PR will be automatically updated.

cdcseacave:
sorry, I saw your comment just after I pushed the new commit, hope this works

pmoulon:
Since this binary export a scene without boostless serialization I have some doubt about the behavior in OpenMVG.
I think OpenMVS will not be able to read it; since by default if boost is detected it only read BOOST serialized scene....
I think we should add a _USE_BOOST_SERIALIZATION in order to make program_options still usable, but boost serialization only a compile time option.

pmoulon:
@cdcseacave 
I have tested the interface and it works fine with latest master/develop OpenMVS.
Thanks you!
Perhaps later we will improve the export in order to allow multithreaded image un-distortion...

Hoping that the interface will remains stable from version to version ;-)

PR #513, opened 2016-03-29
[apps] Feature masking

### LINKED ISSUE #440: Masking before SIFT
I saw masking-ability in the code of the OpenCV-feature-detecting but it is somewhat undocumented. Can you provide an example of how to run feature-detection while leaving out masked areas?

ORNis:
Close #440.
this PR adds a global mask and an individual mask mechanism. The individual mask takes precedence over the global one.
this PR is only for code review for now. 

* [x] Add the mechanism to OpenCV_ComputeFeatures
* [x] Add some documentation
* [x] Ignore "mask files" in SfMInit_ImageListing
* [x] Try to reproduce the issue I mentioned in #440
* [x] Masking rules: a point is masked iff (mask(pointposition)==0)

### DIFF
--- a/src/software/SfM/main_ComputeFeatures.cpp
+++ b/src/software/SfM/main_ComputeFeatures.cpp
@@ -56,7 +56,6 @@ int main(int argc, char **argv)
   std::string sImage_Describer_Method = "SIFT";
   bool bForce = false;
   std::string sFeaturePreset = "";
-  bool bUseMask = false;
 
   // required
   cmd.add( make_option('i', sSfM_Data_Filename, "input_file") );
@@ -66,7 +65,6 @@ int main(int argc, char **argv)
   cmd.add( make_option('u', bUpRight, "upright") );
   cmd.add( make_option('f', bForce, "force") );
   cmd.add( make_option('p', sFeaturePreset, "describerPreset") );
-  cmd.add( make_option('k', bUseMask, "mask") );
 
   try {
       if (argc == 1) throw std::string("Invalid command line parameter.");
@@ -88,7 +86,6 @@ int main(int argc, char **argv)
       << "   NORMAL (default),\n"
       << "   HIGH,\n"
       << "   ULTRA: !!Can take long time!!\n"
-      << "[-k|--mask] use mask to filter regions\n"
       << std::endl;
 
       std::cerr << s << std::endl;
@@ -102,8 +99,7 @@ int main(int argc, char **argv)
             << "--describerMethod " << sImage_Describer_Method << std::endl
             << "--upright " << bUpRight << std::endl
             << "--describerPreset " << (sFeaturePreset.empty() ? "NORMAL" : sFeaturePreset) << std::endl
-            << "--force " << bForce << std::endl
-            << "--mask " << bUseMask << std::endl;
+            << "--force " << bForce << std::endl;
 
 
   if (sOutDir.empty())  {
@@ -219,7 +215,7 @@ int main(int argc, char **argv)
     Image<unsigned char> imageMask;
 
     const std::string sGlobalMask_filename = stlplus::create_filespec(sfm_data.s_root_path, "mask.png");
-    if(bUseMask && stlplus::file_exists(sGlobalMask_filename))
+    if(stlplus::file_exists(sGlobalMask_filename))
       ReadImage(sGlobalMask_filename.c_str(), &globalMask);
 
     C_Progress_display my_progress_bar( sfm_data.GetViews().size(),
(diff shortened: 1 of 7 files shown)


src/software/SfM/main_ComputeFeatures.cpp
```diff
@@ -211,6 +216,13 @@ int main(int argc, char **argv)
   {
     system::Timer timer;
     Image<unsigned char> imageGray;
+    Image<unsigned char> globalMask;
+    Image<unsigned char> imageMask;
```
pmoulon: Perhaps we can use only one image names imageMask.

src/software/SfM/main_ComputeFeatures.cpp
```diff
@@ -231,9 +243,22 @@ int main(int argc, char **argv)
...
 
+        Image<unsigned char> * mask = nullptr;
+
+        const std::string sImageMask_filename = stlplus::create_filespec(sfm_data.s_root_path,
+         stlplus::basename_part(sView_filename), "_mask.png");
+
+        if(bUseMask && stlplus::file_exists(sImageMask_filename))
+          ReadImage(sImageMask_filename.c_str(), &imageMask);
+
+        if(globalMask.size() == imageGray.size())
```
pmoulon: This test will be simplified if we use only one image

src/software/SfM/main_ComputeFeatures.cpp
```diff
@@ -231,9 +243,22 @@ int main(int argc, char **argv)
         if (!ReadImage(sView_filename.c_str(), &imageGray))
           continue;
 
+        Image<unsigned char> * mask = nullptr;
+
+        const std::string sImageMask_filename = stlplus::create_filespec(sfm_data.s_root_path,
+         stlplus::basename_part(sView_filename), "_mask.png");
+
+        if(bUseMask && stlplus::file_exists(sImageMask_filename))
```
pmoulon: Global mask does not have the priority here (perhaps we can change the test to)
if bUseMask && globalMask.Width() > 0 && file exists...

src/software/SfM/main_ComputeFeatures.cpp
```diff
@@ -231,9 +243,22 @@ int main(int argc, char **argv)
...
+          ReadImage(sImageMask_filename.c_str(), &imageMask);
+
+        if(globalMask.size() == imageGray.size())
+          mask = &globalMask;
+        if(imageMask.size() == imageGray.size())
```
pmoulon: Use an else in order to restore global mask priority?

pmoulon:
First: Thanks for the PR and the request for a CR.
=> I made a first fast review:
I think the code looks fine, the local mask can take the ownership to have a custom mask for only some images and it's a cool feature.
We have a add the doc & update the OpenCV_ComputeFeatures sample and perhaps add a unit test to ensure that feature in 'dark' area of a mask are not kept

ORNis:
Thanks a lot Pierre for your CR! The individual/global mechanism was your idea.
I think we should modify main_SfMInit_ImageListing to ignore mask files too and also add the option/flag to main_ComputeStructureFromKnownPoses. What do you think about this ?

pmoulon:
For main_ComputeStructureFromKnownPoses it must be direct, since features that have been masked will not be saved on disk; so later on they cannot be used.

Good point for main_SfMInit_ImageListing => we must check that the substring mask.png or _mask.png cannot be found in the filename part of the view.

ORNis:
Before going ahead, I'm not sure of two things:
- Add a flag (i.e. -k) to binaries that take advantages of masks. Maybe it's better if the use of masks is implicit since we use strong naming conventions. User can move/delete masks images if he do not want to use them anymore.
- Unit test seems a bit tricky since the filtering is feature dependent and encapsulated in the describe method. 

The good news is that after many real-life tests, I wasn't able to reproduce the strange behavior I mentioned in #440.

ORNis:
I think I'm done with this PR.
Please tell me if you see other room for improvements.
Don't hesitate to make changes to the documentation, it's very raw material, my English is really not that good.
thanks.

src/software/SfM/main_ComputeFeatures.cpp
```diff
@@ -219,7 +215,7 @@ int main(int argc, char **argv)
     Image<unsigned char> imageMask;
 
     const std::string sGlobalMask_filename = stlplus::create_filespec(sfm_data.s_root_path, "mask.png");
```
pmoulon: Here you are using the main image path => -i
But in the doc you was mentioning the path of the sfm_data file => -o
Which one do you think is the better?

ORNis: Initially the SfM container path because sfm_data is the most "global" thing we could have :). But you can choose something else ultimately.

pmoulon: Ok, will update the compute features binaries and the PR will be merged.

pmoulon:
Merged https://github.com/openMVG/openMVG/commit/2be24ea19769482c3b02adcaf3f4364d31d1942b

PR #527, opened 2016-04-19
Code cleanup and C++11

rperrot:
This is an attempt to clean some part of the code and making it more C++11 like. 

- Code cleanup fix : 
* some typo (acrancac -> acransac namespace)
* some end namespace comments 

- Code cleanup change some of the for loops on iterators to have range based for loop 
- Code cleanup use C++11 OOP construct (=default, override and so on)

### DIFF
--- a/src/openMVG/cameras/Camera_IO.hpp
+++ b/src/openMVG/cameras/Camera_IO.hpp
@@ -24,15 +24,22 @@ static bool save(
   const PinholeCamera & cam)
 {
   if (stlplus::extension_part(scameraFile) != "bin")
-    return false;
+  {
+    return false;    
+  }
+  else 
+  {
+    const Mat34 & PMat = cam._P;
+    std::ofstream file( scameraFile.c_str(), std::ios::out|std::ios::binary);
+    if( ! file )
+      return false ; 
+    
+    file.write(reinterpret_cast<const char*>(PMat.data()),(std::streamsize)(3*4)*sizeof(double));
 
-  const Mat34 & PMat = cam._P;
-  std::ofstream file( scameraFile.c_str(), std::ios::out|std::ios::binary);
-  file.write((const char*)PMat.data(),(std::streamsize)(3*4)*sizeof(double));
-
-  bool bOk = (!file.fail());
-  file.close();
-  return bOk;
+    const bool bOk = (!file.fail());
+    file.close();
+    return bOk;    
+  }
 }
 
 /// Load a Pinhole camera from a file (read 12 doubles saved in binary values)
@@ -51,7 +58,7 @@ static bool load(
       return false;
     }
     val.resize(12);
-    in.read((char*)&val[0],(std::streamsize)12*sizeof(double));
+    in.read(reinterpret_cast<char*>(&val[0]),(std::streamsize)12*sizeof(double));
     if (in.fail())  {
       val.clear();
     }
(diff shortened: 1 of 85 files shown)


src/openMVG/cameras/Camera_IO.hpp
```diff
@@ -24,15 +24,22 @@ static bool save(
...
-  std::ofstream file( scameraFile.c_str(), std::ios::out|std::ios::binary);
-  file.write((const char*)PMat.data(),(std::streamsize)(3*4)*sizeof(double));
-
-  bool bOk = (!file.fail());
-  file.close();
-  return bOk;
+  {
+    return false;    
+  }
+  else
```
simogasp: i think we can avoid the `else` branch here, it improves the code readability.

src/openMVG/matching_image_collection/Pair_Builder.hpp
```diff
@@ -103,10 +103,9 @@ static bool savePairs(const std::string &sFileName, const Pair_Set & pairs)
       << "savePairs: Impossible to open the output specified file: \"" << sFileName << "\"." << std::endl;
     return false;
   }
-  for (Pair_Set::const_iterator iterP = pairs.begin();
-    iterP != pairs.end(); ++iterP)
-  {
-    outStream << iterP->first << ' ' << iterP->second << '\n';
+  for ( const auto & cur_pair : pairs ) 
+    {
```
pmoulon: Indentation

src/openMVG/robust_estimation/rand_sampling.hpp
```diff
@@ -35,10 +35,11 @@
 namespace openMVG {
 namespace robust{
 
+// rperrot : is this legal ? why using an anonymous namespace here ? 
 namespace
 {
   std::default_random_engine random_generator;
```
pmoulon: In order to do not declare it as static. I added the thread_local close too in the develop branch

src/openMVG/robust_estimation/robust_estimator_ACRansac.hpp
```diff
@@ -49,7 +49,7 @@
 namespace openMVG {
 namespace robust{
 
-namespace acrancac_nfa_internal {
+namespace acransac_nfa_internal {
```
pmoulon: Typed too fast. Worth it ;-)

src/openMVG/robust_estimation/rand_sampling.hpp
fabiencastan: It doesn't make sense to use anonymous namespaces in headers. It will duplicate the symbols in each cpp that include this header. If you really need a global variable you need to declare it in static in the cpp file and in extern in the header.

Here, you can declare it as a static variable directly inside the function.

src/openMVG/cameras/Camera_IO.hpp
rperrot: I agree, I mainly use clang-tidy tool and it produces a warning here if no else branch is present. I think it's a false positive and I did not make any modification like that elsewhere with the same warning.

src/openMVG/robust_estimation/rand_sampling.hpp
rperrot: Thread locale does not compile on OSX, just tested yet with current develop branch  : 

```
openMVG/src/openMVG/robust_estimation/rand_sampling.hpp:40:3: error: thread-local
      storage is not supported for the current target
  thread_local std::default_random_engine random_generator;
```

pmoulon: Curious that it is not supported since this is a c++11 feature http://en.cppreference.com/w/cpp/language/storage_duration
Perhaps will move the variable declaration to the function where it is used std::default_random_engine random_generator; Don't know the speed impact.

rperrot: As stated in the clang doc, it's a libc++ issue.

pmoulon:
Merged :+1:

PR #528, opened 2016-04-20
Fix clang compilation issue with std::move on const objects

rperrot:
Here is a quick fix for compilation on recent clang that seems stricter than previous version when using std::move on const objects. Since destination object doesn't have const attribute, this generate an error. 

Here is an example of the error generated without the fix : 

```
openMVG/src/openMVG/matching_image_collection/GeometricFilter.hpp:94:55: error: 
      binding value of type 'const pair<[2 * ...]>' to reference to type 'pair<[2 * ...]>' drops 'const' qualifier
          _map_GeometricMatches.insert(std::make_pair(current_pair, std::move(putative_inliers)));
                                                      ^~~~~~~~~~~~
```

### DIFF
--- a/src/openMVG/matching_image_collection/Cascade_Hashing_Matcher_Regions_AllInMemory.cpp
+++ b/src/openMVG/matching_image_collection/Cascade_Hashing_Matcher_Regions_AllInMemory.cpp
@@ -109,7 +109,7 @@ void Match
     const size_t dimension = regionsI.DescriptorLength();
 
     Eigen::Map<BaseMat> mat_I( (ScalarT*)tabI, regionsI.RegionCount(), dimension);
-    const HashedDescriptions hashed_description = cascade_hasher.CreateHashedDescriptions(mat_I,
+    HashedDescriptions hashed_description = cascade_hasher.CreateHashedDescriptions(mat_I,
       zero_mean_descriptor);
 #ifdef OPENMVG_USE_OPENMP
     #pragma omp critical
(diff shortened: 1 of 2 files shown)


src/openMVG/matching_image_collection/Cascade_Hashing_Matcher_Regions_AllInMemory.cpp
```diff
@@ -144,7 +144,7 @@ void Match
 #endif
     for (int j = 0; j < (int)indexToCompare.size(); ++j)
     {
-      const size_t J = indexToCompare[j];
+      size_t J = indexToCompare[j];
```
pmoulon: Strage that J must be non const, since it is just used as a look-up index later on

rperrot: if you have : 

```
const size_t J
```

using std::move on it produces this type : 
```
const size_t &&
```

but that type can't be used later in std::make_pair : 

```
openMVG/src/openMVG/matching_image_collection/Cascade_Hashing_Matcher_Regions_AllInMemory.cpp:212:63: error: 
      binding value of type 'const size_t' (aka 'const unsigned long') to reference to type 'unsigned long' drops 'const' qualifier
          map_PutativesMatches.insert( make_pair( make_pair(I,J), std::move(vec_putative_matches) ));
```

pmoulon: ok, Thx

rperrot: Moreover, it seems that in c++11 standard there's no const arguments on std::make_pair : 

http://en.cppreference.com/w/cpp/utility/pair/make_pair

PR #543, opened 2016-04-29
Removed constexpr

mitjap:
`constexpr` is only supported in VC14 (VS2015). Another solution would be to use definitions to check for compilers older then VC14 (VS2015)


pmoulon:
I will keep the constexprt.
Feel free to make a PR with the ifdef for VS

PR #555, opened 2016-05-10
Develop parallel export images

hcjghr:
Hi Pierre,

this pull request adds the possibility of the parallel export of undistorted images  (following the same approach as in _main_ComputeFeatures.cpp_). Additionally I added the option of exporting only the images that have been successfully reconstructed (in some my cases that is preferred).

Klemen


src/software/SfM/main_ExportUndistortedImages.cpp
```diff
@@ -19,14 +20,27 @@ using namespace openMVG::sfm;
...
   std::string sSfM_Data_Filename;
   std::string sOutDir = "";
+  bool bExportOnlyReconstructedViews = false;
+#ifdef OPENMVG_USE_OPENMP
+  int iNumThreads = 0;
+#endif
 
   cmd.add( make_option('i', sSfM_Data_Filename, "sfmdata") );
   cmd.add( make_option('o', sOutDir, "outdir") );
+  cmd.add( make_option('r', bExportOnlyReconstructedViews, "exportOnlyReconstructed") );
```
pmoulon: perhaps we can use the make_switch -R and check that the command have been used with cmd.used(option)
see https://github.com/openMVG/openMVG/blob/master/src/software/SfM/main_ConvertSfM_DataFormat.cpp#L65

src/software/SfM/main_ExportUndistortedImages.cpp
```diff
@@ -48,21 +66,41 @@ int main(int argc, char *argv[]) {
...
+    
+    #ifdef OPENMVG_USE_OPENMP
+    const unsigned int nb_max_thread = omp_get_max_threads();
+    #endif
+    
+#ifdef OPENMVG_USE_OPENMP
+    omp_set_num_threads(iNumThreads);
+    #pragma omp parallel for schedule(dynamic) if(iNumThreads > 0) private(image,image_ud)
+#endif
+    for(int i = 0; i < sfm_data.views.size(); ++i)
```
pmoulon: for" "(

src/software/SfM/main_ExportUndistortedImages.cpp
```diff
@@ -48,21 +66,41 @@ int main(int argc, char *argv[]) {
...
     {
-      const View * view = iter->second.get();
+#ifdef OPENMVG_USE_OPENMP
+      if(iNumThreads == 0) omp_set_num_threads(nb_max_thread);
```
pmoulon: if" "(

src/software/SfM/main_ExportUndistortedImages.cpp
```diff
@@ -48,21 +66,41 @@ int main(int argc, char *argv[]) {
...
+#endif
+      Views::const_iterator iterViews = sfm_data.views.begin();
+      std::advance(iterViews, i);     
+    
+      const View * view = iterViews->second.get();
+      // Check if the view is in reconstruction
+      if(bExportOnlyReconstructedViews && !sfm_data.IsPoseAndIntrinsicDefined(view))
```
pmoulon: if  (

src/software/SfM/main_ExportUndistortedImages.cpp
```diff
@@ -48,21 +66,41 @@ int main(int argc, char *argv[]) {
...
+        continue;
+        
       bool bIntrinsicDefined = view->id_intrinsic != UndefinedIndexT &&
```
pmoulon: const bool

pmoulon:
Thanks.
I have made some minor comments in the code review.
The only problem I see is that the program can crash if the thread allocate more memory than the machine can provide.
That's why the default option for computeFeatures is thread count to 0 and the expoert user then can configure the thread count depending of it's project

pmoulon:
Thanks you @hcjghr, Your PR is now merged! (Sorry for the late merge)
Hoping this is the first of a long series!

PR #592, opened 2016-07-13
Integrating KFAST - 60% faster than any other known FAST

mdaiter:
Could I integrate this version of FAST into OpenMVG? I noticed that there isn't a solid version of FAST integrated yet, and I'd love to take the lead on integrating this into the main tree of OpenMVG.

KFAST was developed by @komrad36 using AVX2.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
+[submodule "src/openMVG/features/fast"]
+	path = src/openMVG/features/fast
+	url = https://github.com/komrad36/KFAST_OpenMVG.git
 [submodule "src/dependencies/glfw"]
 	path = src/dependencies/glfw
 	url = https://github.com/elmindreda/glfw.git
--- a/src/openMVG/features/CMakeLists.txt
+++ b/src/openMVG/features/CMakeLists.txt
@@ -17,14 +17,23 @@ file(GLOB_RECURSE REMOVEFILESUNITTEST *_test.cpp)
 #Remove the future main files
 list(REMOVE_ITEM features_files_sources ${REMOVEFILESUNITTEST})
 
+file(GLOB_RECURSE REMOVEFILESFASTCPP fast/*.cpp)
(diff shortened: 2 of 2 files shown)


pmoulon:
Hi,
Thank you @mdaiter & @komrad36 for the PR.

1. We must take care that it will not be in conflict with the WIP there https://github.com/openMVG/openMVG/tree/develop_VO/src/third_party/fast

2. Regarding the integration, can you provide the required compilation flag?
 - AVX2 & SSE4...

3. The integration must be clean, I mean that we must check in cmake if we can use the Kfast or the fast lib and use the appropriate one.

mdaiter:
Hey @pmoulon ,
Thanks for the review.
1. The branch seems to pass all checks through the compilation phase. However, we can change this back to the original project name KFAST if you'd like, to ensure avoiding any future conflicts.
2. All compilation flags required are included in the bundled CMakeLists.txt file that comes with the project. AVX2 and SSE4 are already included in the compilation flags.
3. Do you mean that the appropriate library, depending on the capabilities of the processor, is compiled and included in the project?
Thanks again.

pmoulon:
What I mean is that in the final version we must include a similar interface for FAST & KFAST.
In order to compile FAST or KFAST iff supported by the architecture & compiler.
So yes, I perfer the way 3=> appropriate library is chosen depending of the capabilities of the processor (at compilation time)

pmoulon:
@mdaiter I have merged the WIP develop_VO branch in the develop trunk.
So you will have a good base to work on integrating kfast and switch between fast or kfast if AVX2 is supported by the computer or not.

.gitmodules
```diff
@@ -1,3 +1,6 @@
+[submodule "src/openMVG/features/fast"]
+	path = src/openMVG/features/fast
```
pmoulon: Would be cool if we keep the Kfast name for the path
Ideally we can think about integrating directly the code, or fork the repo there https://github.com/openMVG-thirdparty in order to never loose the external git repo if a user decide to remove it's account.

PR #599, opened 2016-07-27
Adding more opencv feature detectors for comparison

squeakus:
Hi, I have added surf, orb and brisk from the opencv library so that feature descriptors can be compared. My apologies if I have done this incorrectly but it is my first pull request and wanted to give something back to your excellent project.


donlk:
If more OpenCV based feature detectors are to be integrated, perhaps it would be reasonable to just create a factory for them (or just wrap around the one in OpenCV), as the only unique values among theseare their type and descriptor size. Otherwise the redundancy will soon grow unbearable along with maintenance cost.

squeakus:
okay, let me give it a bash.

src/openMVG/exif/sensor_width_database/sensor_width_camera_database.txt
```diff
@@ -656,6 +657,7 @@ Contax TVS Digital;7.11
 Contax N Digital;36
 DJI FC300S;6.16
 DJI FC300X;6.16
+#FC300X;6.16
```
fabiencastan: Why `#`?

pmoulon: We will ignore this file for any PR, since we want one feature or one enhancement per PR.

fabiencastan:
Hi,

Please create your PR to `develop` and not directly to `master` which is dedicated for releases.

It could be great to move all image_describers related to openCV directly into 'openMVG/feature' with the `#ifdef HAVE_OPENCV` instead of keeping all this code into the `main*.cpp`. So we can have only one `main_ComputeFeatures.cpp` with all describer types.
@pmoulon What do you think?

pmoulon:
The goal is not to integrate openCV stuff inside OpenMVG, since OpenMVG have it's own detector/descriptor.
The initial goal was to show that you can plug easily external image describer (a kind of tutorial/sample)

src/openMVG/exif/sensor_width_database/sensor_width_camera_database.txt
squeakus: yes, ignore that. I was testing a script that will guess a focal length if it is not in the repo.

squeakus:
@donlk I have tried abstracting it into a factory but as the different descriptors are in different parts of the opencv library it makes it difficult. I tried overloading the image_describer.reset function but grep is not finding it. I am still new to the code so still trying to understand the underlying logic. 

@pmoulon if you want to delete this pull request that is no problem. I just wanted to make my code available, thanks for providing the examples and hopefully I will have something better to contribute when I understand the code more!

pmoulon:
@squeakus You're welcome. It's nice to share your code experiments so that anyone can make constructive comments on it.
It's seems that it's not easy to find a generic way for the integration.
Let's continue the discussion about this PR a bit.

BTW, I let the PR open, so that anyone interested by doing a similar things can read and see your modification.

pmoulon:
BTW, @squeakus do you have interesting result to share in term of speed and/or accuracy when you tested the openCV image_describer interface?

squeakus:
I am still only doing early stage testing but so far the time taken is
almost the same as when called directly in opencv. I am halfway through
writing the paper and will send on the results once everything is graphed!

donlk:
@squeakus Yes, OpenCV has a lot do be desired when it comes to module consistency. They are still heavily developing a generic acceleration interface that would presumably make functions use the best backend for the hardware, but its still not finished. If you want to use cuda based feature detection for example you need to switch base class and namespace altogether.
Therefore i think the best option would be just to provide a wrapper for the detectors/descriptors already  integrated into a factory module here:
https://github.com/opencv/opencv/blob/master/modules/features2d/include/opencv2/features2d.hpp
With this you only need to wrap around cv::FeatureDetector and cv::DesciptorExtractor in your factory and modify only the module type through the appropriate create(...) methods.

squeakus:
@donlk thanks for sending me on that info. I will have a day or two next
week free to code and I will try implementing it then.

tomas789:
@pmoulon Are there any intentions with this PR? I would very appreciate having this in core repository. It is very valuable to have as many feature detectors/descriptors available directly. Is there any strong reason for not to including it?

PR #600, opened 2016-07-28
[build/continuous integration] Add clang compiler to Travis CI

pmoulon:
- Fix some clang compilation warning,
- Fix a SfM_Localizer error detected by clang,
- [sfm/pipelines/sequential] Do not optimize too early the intrinsic when a resection is performed.

### DIFF
--- a/src/openMVG/sfm/pipelines/sequential/sequential_SfM.cpp
+++ b/src/openMVG/sfm/pipelines/sequential/sequential_SfM.cpp
@@ -1025,9 +1025,11 @@ bool SequentialSfMReconstructionEngine::Resection(const size_t viewIndex)
           return false;
       }
     }
+    const bool b_refine_pose = true;
+    const bool b_refine_intrinsics = false;
     if(!sfm::SfM_Localizer::RefinePose(
-      optional_intrinsic.get(), pose,
-      resection_data, true, b_new_intrinsic))
+        optional_intrinsic.get(), pose,
+        resection_data, b_refine_pose, b_refine_intrinsics))
     {
       return false;
     }
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,25 @@
 
 language: cpp
 
-compiler:
-  - gcc
-  #- clang: #Consider clang later, since cereal does not build fine on the clang CI version
-  #  - "3.3"
+matrix:
+  include:
+    - compiler: gcc
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
+            - g++-4.8
+      env: COMPILER=g++-4.8
+    - compiler: clang
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+            - llvm-toolchain-precise-3.7
+          packages:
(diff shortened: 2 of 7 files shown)


src/openMVG/sfm/pipelines/sequential/sequential_SfM.cpp
```diff
@@ -1027,7 +1027,7 @@ bool SequentialSfMReconstructionEngine::Resection(const size_t viewIndex)
     }
     if(!sfm::SfM_Localizer::RefinePose(
       optional_intrinsic.get(), pose,
-      resection_data, true, b_new_intrinsic))
+      resection_data, true, false))
```
simogasp: @pmoulon Just out of curiosity, do you notice a difference between an early optimization (the root of all evil :-D) and a delayed (intrinsics) optimization during the BA? I was wondering myself what is the impact of having a "good/optimized" solution before doing the triangulation of the new points, and use a "first guess" solution (with no distortion coefficients estimated)  to do triangulation and then optimize/refine everything in the BA. 
One thing that comes to mind, is that with delayed optimization maybe the estimation of the distortion coefficients could be better, because *in theory* you can cover more regions of the image with the new points (while sometimes when you do resection the points could be in one side/region of the image, and thus if u refine just after that you may bias a bit the estimation of the distortion (or not, i really don't know)). But of course it depends on images etc.
On the other hand it's true that triangulation without undistorted points may introduce some (or a lot of noise), especially if the points are far from the image center. And I don't know if BA can fix that later.
Did you notice some effect like that in your experiments?

src/openMVG/sfm/pipelines/sequential/sequential_SfM.cpp
```diff
@@ -1027,7 +1027,7 @@ bool SequentialSfMReconstructionEngine::Resection(const size_t viewIndex)
     }
     if(!sfm::SfM_Localizer::RefinePose(
```
simogasp: what about "wasting" 2 booleans like that
```c++
const bool b_refine_pose = true;
const bool b_refine_intrinsics = false;
```
and use it instead of explicit `true` and `false`. That way whoever reads, can understand what is happening without having to jump to the definition of `RefinePose`.

src/openMVG/sfm/pipelines/sequential/sequential_SfM.cpp
pmoulon: I think that delayed optimization will improve the results too. But I also agree that it will depends from dataset to dataset.
Since often you can use pre-calibrated camera, you will not have those problems.
But when you want to make some auto-calibration (from a guess) it's better to perform delayed refinement in order to maximize the observations per images in order to have an enhanced set of image observation.

src/openMVG/sfm/pipelines/sequential/sequential_SfM.cpp
pmoulon: I agree. Good catch, it will help to make things more explicit ;-).
The cons, is that perhaps more user will play with hidden parameters since they are more explicit :-(
Thx, I will update that soon.

PR #611, opened 2016-08-18
fixed link to MVE's TexRecon specification

mitjap:


pmoulon:
Thank you @mitjap. Will handle that in the coming days

PR #622, opened 2016-09-09
Develop fast resection

hcjghr:
Hi @pmoulon,

As I was reconstructing a model with approx. 2000 views I noticed that the augmentation of already reconstructed landmarks and triangulation of new ones in the resection method took significant amount of time (in my case after  around 1000 views were added only this part of resection took 13s).  

The reason was that the procedure took each already reconstructed view and found all the tracks that are in common with the newly added view. These tracks were then searched to find the tracks that were already reconstructed (and so the new observations could be added) or new landmark can be created through the triangulation process between the steps. As the number of reconstructed views increased this took longer and longer.

In the method I propose the procedure takes all the tracks that are seen in the newly added view and check each of them to see if it was already reconstructed (only add the observation from the current view)  and if not checks all the views that see each track. From the views that have already been reconstructed it tries to triangulate a new landmark and add the rest as observations.

In the same dataset this reduced the computational time of this step to approx 1.5s.

Is this a valid thinking or am I missing an important fact for which reason the "old" procedure should be used?

best

### DIFF
--- a/src/openMVG/sfm/pipelines/sequential/sequential_SfM.cpp
+++ b/src/openMVG/sfm/pipelines/sequential/sequential_SfM.cpp
@@ -1096,8 +1096,7 @@ bool SequentialSfMReconstructionEngine::Resection(const size_t viewIndex)
 		// If the track was already reconstructed
 		if (sfm_data_.structure.count(trackId) != 0)
 		{
-		  // 3D point triangulated before, only add image observation if needed
-		  {
+			// 3D point triangulated before, only add image observation if needed
 			Landmark & landmark = sfm_data_.structure[trackId];
 			if (landmark.obs.count(I) == 0)
 			{
@@ -1107,7 +1106,6 @@ bool SequentialSfMReconstructionEngine::Resection(const size_t viewIndex)
 					landmark.obs[I] = Observation(xI, track.at(I));
 				}
 			}
-		  }
 		}
 		else
 		{
@@ -1122,10 +1120,12 @@ bool SequentialSfMReconstructionEngine::Resection(const size_t viewIndex)
 				if(J!=I && valid_views.count(J) != 0 )
 				{
 					// If successfuly triangulated add the observation from J view
-					if (sfm_data_.structure.count(trackId) != 0){
+					if (sfm_data_.structure.count(trackId) != 0)
+					{
 						new_track_observations_valid_views.insert(J); 
 					}
-					else{
+					else
+					{
 						const View * view_J = sfm_data_.GetViews().at(J).get();
 						const IntrinsicBase * cam_J = sfm_data_.GetIntrinsics().at(view_J->id_intrinsic).get();
 						const Pose3 pose_J = sfm_data_.GetPoseOrDie(view_J);
@@ -1159,7 +1159,8 @@ bool SequentialSfMReconstructionEngine::Resection(const size_t viewIndex)
 							landmark.obs[I] = Observation(xI, track.at(I));
 							landmark.obs[J] = Observation(xJ, allViews_of_track.at(J));
 						} // 3D point is valid
-						else{
+						else
+						{
 							// We mark the view to add the observations once the point is triangulated
 							new_track_observations_valid_views.insert(J);  
 						}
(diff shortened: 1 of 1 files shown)


src/openMVG/sfm/pipelines/sequential/sequential_SfM.cpp
```diff
@@ -1076,124 +1076,121 @@ bool SequentialSfMReconstructionEngine::Resection(const size_t viewIndex)
   // G. Triangulate new possible 2D tracks
   // List tracks that share content with this view and add observations and new 3D track if required.
   {
-    // For all reconstructed images look for common content in the tracks.
+	// Get information of new view
+    const size_t I =(IndexT)viewIndex;
```
pmoulon: const IndexT I = viewIndex;

src/openMVG/sfm/pipelines/sequential/sequential_SfM.cpp
```diff
@@ -1076,124 +1076,121 @@ bool SequentialSfMReconstructionEngine::Resection(const size_t viewIndex)
...
-          << "\n--Triangulated 3D points [" << I << "-" << J << "]:"
-          << "\n\t#Track extented: " << extented_track
-          << "\n\t#Validated/#Possible: " << new_added_track << "/" << new_putative_track
-          << "\n\t#3DPoint for the entire scene: " << sfm_data_.GetLandmarks().size() << std::endl;
-      }
-*/
-    }
+    // Go through each track in the new view and either add observations or new 3D points
+    for (const std::pair< size_t, tracks::submapTrack >& trackIt : map_tracksCommon)
+	{
```
pmoulon: alignment issue, I can fix them later on. tab = 2 space in OpenMVG

src/openMVG/sfm/pipelines/sequential/sequential_SfM.cpp
```diff
@@ -1076,124 +1076,121 @@ bool SequentialSfMReconstructionEngine::Resection(const size_t viewIndex)
...
+				// If view is valid try triangulation
+				if(J!=I && valid_views.count(J) != 0 )
+				{
+					const View * view_J = sfm_data_.GetViews().at(J).get();
+					const IntrinsicBase * cam_J = sfm_data_.GetIntrinsics().at(view_J->id_intrinsic).get();
+					const Pose3 pose_J = sfm_data_.GetPoseOrDie(view_J);
+					const Vec2 xJ = features_provider_->feats_per_view.at(J)[allViews_of_track.at(J)].coords().cast<double>();
+
+					// If successfuly triangulated add the observation from J view
+					if (sfm_data_.structure.count(trackId) != 0){
```
pmoulon: I think this IF part can be simplified to
```
new_track_observations_valid_views.insert(J);
```
And so the point observation will be added by your loop below.
It will make the code simpler.

pmoulon:
@hcjghr Thank you for this PR!

For sure the existing loop was not the best for the complexity. This loop was due to some tests I made in the past when the track structure was dynamic during the reconstruction. Since now that it is static there is not risk to use your method and make this loop shorter run over a reduced set of view ids!

**Here some questions and comments:**
- *OpenMP:* I saw that you have removed the parallelism in the tracks loop. Does it was wanted or does the loop is faster without using OpenMP?
- *Code style:* Please use 2 spaces tabs and brace on new line. Thanks.
- *Algorithm:* I made a suggestion to avoid some code duplication in the first IF

We can in a second time step go further in the optimization:
- We can make 3D-2d track correspondences looking faster:
  - Actually the total structure observation is parsed each time.
  - But we can pre-compute an array of track Id per view and so looking for 3D-2D association will be just a matter of set intersection (that should be run very fast since of the nature set (sorted arrays)).

=> We can take care of that in a second PR.

PS:
When I have **medium/huge** dataset I often use Global SfM that is far faster than the Incremental pipeline (due to the fact of its non iterative process).

src/openMVG/sfm/pipelines/sequential/sequential_SfM.cpp
```diff
@@ -1101,11 +1101,11 @@ bool SequentialSfMReconstructionEngine::Resection(const size_t viewIndex)
 			Landmark & landmark = sfm_data_.structure[trackId];
```
pmoulon: There is still 4 spaces indentation ;-)

src/openMVG/sfm/pipelines/sequential/sequential_SfM.cpp
```diff
@@ -1076,124 +1076,116 @@ bool SequentialSfMReconstructionEngine::Resection(const size_t viewIndex)
   // G. Triangulate new possible 2D tracks
   // List tracks that share content with this view and add observations and new 3D track if required.
   {
-    // For all reconstructed images look for common content in the tracks.
+	// Get information of new view
```
pmoulon: indentation

pmoulon:
Merged:
https://github.com/openMVG/openMVG/commit/9b7091d3e83e9595fca760e9e805f399c788d74e
and updated to be even simpler:
https://github.com/openMVG/openMVG/commit/fcbc0779f6e1c85b3c80f3cbd9424c11e0f6d8a8

Thank you @hcjghr :+1:

pmoulon:
Forgot to mention that on large scene this tiny update can make the process up to 30% faster. :tada:

Issue #618, opened 2016-08-26
Adding views to an existing reconstruction


sebastienchappuis:
Hello,

First, thank you for this awesome software and for making it opensource !

I was wondering if there was a way to add cameras/views to an already reconstructed dataset. 
Basically, if I have a reconstruction of a 1000 images and I want to add a 100 images to the scene, is it possible to do so without having to recompute the first 1000 camera resections ? 
My question is more relevant for sequential reconstruction than global reconstruction I guess.

Thanks !

pmoulon:
In order to localize some new images inside a reconstruction you can use the Localization module.
See the ```$ openMVG_main_SfM_Localization``` binary.
It will try to find corresponding 2D-3D matches to the existing scene structure and to compute camera resection (it will not add new 3D points).
https://openmvg.readthedocs.io/en/latest/software/localization/localization/

The tool take only one image as input, but you can easily change it to take a series of images or a directory.

Feel free to try it a to continue the discussion and close the discussion once you tested it with success.
Feel also free to help in order to enhance the localization module and it's usage.

sebastienchappuis:
Thanks !

I'll do a few tests and come back to you with my results.

pmoulon:
I forgot to mention that for the moment only a PLY file is logged, I can work more with you next week to extend the binary to add the View to the existing scene and so complete the starting sfm_data with the new poses & intrinsic & view information.

sebastienchappuis:
Ok, so I tested the `Sfm_Localization_Single_3DTrackObservation_Database` localizer used in the `openMVG_main_SfM_Localization` binary. I wrote an alternative binary that takes multiple images as input and outputs a new sfm_data file containing both the original and the newly computed poses/views/intrinsics. All seems to work well. Once I get all the new poses from the localizer I can run something like `openMVG_main_ComputeStructureFromKnownPoses` on the dataset in order to get some reconstruction from all the cameras together.

Now of course, this is not equivalent to resuming sequential reconstruction where it was. In order to get closer to this I would need to add a triangulation step and a BA step after each new camera resection. I will start implementing something in this direction. Maybe you have some advices on how to approach this ?

Thanks for your help !

pmoulon:
@sebastienchappuis Sorry for my late answer.

If you want extend the scene, the task is not trivial, since you have also to update the localization interface at each round.

I would like to do something like this as a first draft:
1. localize the image or a round of images
2. compute to which view the new image(s) are linked => build a list of pair
[...]

PR #629, opened 2016-09-20
Add an executable for localizing multiple images

sebastienchappuis:
As mentionned in issue #618, this is an executable which localizes multiple views in an already existing reconstruction. The executable takes a directory of images as input along with the sfm_data file describing the original reconstruction, and outputs a new sfm_data file that contains all the new views with their poses and intrinsics.

Thank you in advance for your feedback !


src/software/Localization/main_SfM_Localization_MultipleImages.cpp
```diff
@@ -0,0 +1,341 @@
...
+    // Add the view to the sfm_container
+    views[v.id_view] = std::make_shared<View>(v);
+  }
+
+  GroupSharedIntrinsics(sfm_data);
+  
+  std::cout << " Total poses found : " << vec_found_poses.size() << "/" << vec_image_new.size() << endl;
+
+  // Export the found camera position in a ply.
+  const std::string out_file_name = stlplus::create_filespec(sOutDir, "found_pose_centers", "ply");
```
pmoulon: You can use the existing function to do that (save all the camera positions):

FYI:
(**Save** can export scene in PLY and you can choose which information you want to export):
```
Save(
     sfm_data,
      out_file_name .c_str(),
     ESfM_Data(EXTRINSICS)));
```
If you want to save only the new found poses you can use this function:
https://github.com/openMVG/openMVG/blob/master/src/software/SfM/SfMPlyHelper.hpp#L21
```
plyHelper::exportToPly(vec_found_poses, out_file_name);
```

src/software/Localization/main_SfM_Localization_MultipleImages.cpp
```diff
@@ -0,0 +1,341 @@
...
+  // export also as ply
+  if (!Save(
+    sfm_data,
+    stlplus::create_filespec( sOutDir, "sfm_data_expanded.ply" ).c_str(),
+    ESfM_Data(ALL)))
+  {
+    return EXIT_FAILURE;
+  }
+
+  return EXIT_FAILURE;
```
pmoulon: I think the binary must return EXIT_SUCCESS if everything ran fine

src/software/Localization/main_SfM_Localization_MultipleImages.cpp
```diff
@@ -0,0 +1,341 @@
...
+        KRt_From_P(matching_data.projection_matrix, &K, &R, &t);
+
+        const double focal = (K(0,0) + K(1,1))/2.0;
+        const Vec2 principal_point(K(0,2), K(1,2));
+        optional_intrinsic = std::make_shared<cameras::Pinhole_Intrinsic_Radial_K3>(
+          imageGray.Width(), imageGray.Height(),
+          focal, principal_point(0), principal_point(1));
+
+      }
+      sfm::SfM_Localizer::RefinePose
```
pmoulon: we must check the return boolean value, BA can failed to optimize the poses for an unknown reason...

sebastienchappuis:
I did a few corrections. I will soon fuse the two binaries (main_SfM_Localization and main_SfM_Localization_MultipleImages) together.

BTW, Is it still relevant to keep the "demo" part of the main_SfM_Localization executable (from lines 227 to 306)?

pmoulon:
I saw that the CR (Code Review) points are all covered.
You are right, you can merge the two binary (main_SfM_Localization_MultipleImages => main_SfM_Localization)
- [ ] It will require to update the [rst doc](https://github.com/openMVG/openMVG/blob/develop/docs/sphinx/rst/software/localization/localization.rst)
- [ ] the demo mode can be removed
   - This demo mode was an easy way for me to check that the code was working as expected thanks to this mode, since the new camera position must be as close as possible as the original ones.
   - Note: this demo mode can be reproduced within your new binary by removing the set_difference to detect the existing views.

sebastienchappuis:
All right, I fused the two localization executables into one, got rid of the "demo" mode, updated the documentation and fixed a few things !

pmoulon:
Thank you @sebastienchappuis 
I will test it on some scenes and then perform the merge :v:

src/software/Localization/main_SfM_Localization.cpp
```diff
@@ -22,6 +22,17 @@ using namespace openMVG::sfm;
 
 #include <cstdlib>
 
+// Naive function for finding the biggest common root dir from two paths
+std::string FindCommonRootDir(const std::string & dir1, const std::string & dir2)
```
sebastienchappuis: Just a note : I expected to find a function in stlplus that would do this but apparently it does not exist ? I might have missed it though !

pmoulon: You're right, there is not such function.
There is only the function to compare pathes between platforms stlplus::path_compare

pmoulon:
PR merged.
Thank you @sebastienchappuis, hoping this is the start of a long series of future PR ;-)

sebastienchappuis:
> PR merged.
Thank you @sebastienchappuis, hoping this is the start of a long series of future PR ;-)

could very well be :) 
thank you for your help !

PR #638, opened 2016-09-30
Added OpenMP parallelism on image localization

hcjghr:
Hi @pmoulon and @sebastienchappuis,

I've added the OpenMP parallelization to the great implementation of @sebastienchappuis of the multiple image localization. 

Additionally I've added an option to specify a different folder for the features that are extracted from the new images needed to be localized. The reason is that sometimes you want to preserve the original match folder as it is (without the additional files of the feature detection of the new images). If this parameter is not provided it assumes that the features are stored in the same match folder as the ones from the initial reconstruction. 

Klemen

### DIFF
--- a/src/software/Localization/main_SfM_Localization.cpp
+++ b/src/software/Localization/main_SfM_Localization.cpp
@@ -53,6 +53,8 @@ int main(int argc, char **argv)
   std::string sMatchesOutDir;
   std::string sQueryDir;
   double dMaxResidualError = std::numeric_limits<double>::infinity();
+  bool bUseSingleIntrinsics = false;
+  bool bExportStructure = false;
 
 #ifdef OPENMVG_USE_OPENMP
   int iNumThreads = 0;
@@ -65,6 +67,7 @@ int main(int argc, char **argv)
   cmd.add( make_option('q', sQueryDir, "query_image_dir"));
   cmd.add( make_option('r', dMaxResidualError, "residual_error"));
   cmd.add( make_switch('s', "single_intrinsics"));
+  cmd.add( make_switch('e', "export_structure"));
 #ifdef OPENMVG_USE_OPENMP
   cmd.add( make_option('n', iNumThreads, "numThreads") );
 #endif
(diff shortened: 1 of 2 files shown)


src/software/Localization/main_SfM_Localization.cpp
```diff
@@ -235,19 +255,27 @@ int main(int argc, char **argv)
...
 
   int total_num_images = 0;
-  for ( std::vector<std::string>::const_iterator iter_image = vec_image_new.begin();
-    iter_image != vec_image_new.end();
-    ++iter_image)
+  
+#ifdef OPENMVG_USE_OPENMP
+    const unsigned int nb_max_thread = omp_get_max_threads();
+    omp_set_num_threads(iNumThreads);
+    #pragma omp parallel for schedule(dynamic)
```
pmoulon: What about something like
```
const unsigned int nb_max_thread = (iNumThreads == 0) ? 0 : omp_get_max_threads();
omp_set_num_threads(nb_max_thread);
#pragma omp parallel for schedule(dynamic) if(iNumThreads > 0)
```

So you can remove the OPENMVG_USE_OPENMP condition inside the loop

src/software/Localization/main_SfM_Localization.cpp
```diff
@@ -93,7 +93,7 @@ int main(int argc, char **argv)
     << "  contains a single intrinsics and, if so, take this value as intrinsics for the query images.\n"
     << "  (OFF by default)\n"
     << "[-e|--export_structure] (switch) when switched on, the program will also export structure to output sfm_data.\n"
```
pmoulon: Can you motivate why you have chose to set the structure as an option?
Having the structure in all the case can be desired...

hcjghr: As this process only provides a way to localize new images (without extending the current model) the only new information is the pose of the new cameras. If our goal is only to obtain those poses there is no need to export the structure again. This is just to limit the export data size if the models are big and have a lot of landmarks.

pmoulon: Ok, I undertand, but we will have to update the documentation accordingly

pmoulon:
Thx @hcjghr. The PR have been merged.

PR #640, opened 2016-10-02
Added weights for each of the dimension of GPS prior

hcjghr:
Hi @pmoulon,

in this PR i propose a relatively small change to add the ability to put different weights to each of the dimensions of the prior. This would enable to introduce some knowledge about the uncertainty of the estimations for each of the directions. While maybe in the majority of cases the estimates in all 3 are equally good, this might not be the case when we use different sensors for different estimations (in underwater scenarios depth is usually more accurately estimated than the "2D position").  Simple expansion of the weights would enable to add this knowledge into the SfM process. Additionally I added the possibility to initialize the weights through the SfMInit binary.

Best,
Kklemen

### DIFF
--- a/src/software/SfM/main_SfMInit_ImageListing.cpp
+++ b/src/software/SfM/main_SfMInit_ImageListing.cpp
@@ -84,6 +84,35 @@ std::pair<bool, Vec3> checkGPS
   return val;
 }
 
+
+/// Check string of prior weights
+std::pair<bool, Vec3> checkPriorWeightsString
+(
+  const std::string &sWeights
+)
+{
+  std::pair<bool, Vec3> val(true, Vec3::Zero());
+  std::vector<std::string> vec_str;
+  stl::split(sWeights, ';', vec_str);
+  if (vec_str.size() != 3)
+  {
+    std::cerr << "\n Missing ';' character in prior weights" << std::endl;
+    val.first = false;
+  }
+  // Check that all weight values are valid numbers
+  for (size_t i = 0; i < vec_str.size(); ++i)
+  {
+    double readvalue = 0.0;
+    std::stringstream ss;
+    ss.str(vec_str[i]);
+    if (! (ss >> readvalue) )  {
+      std::cerr << "\n Used an invalid not a number character in local frame origin" << std::endl;
+      val.first = false;
+    }
+    val.second[i] = readvalue;
+  }
+  return val;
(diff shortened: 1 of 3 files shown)


src/software/SfM/main_SfMInit_ImageListing.cpp
```diff
@@ -352,6 +391,11 @@ int main(int argc, char **argv)
 
       v.b_use_pose_center_ = true;
       v.pose_center_ = gps_info.second;
+      // prior weights
+      if(prior_w_info.first == true)
```
pmoulon: if (

src/software/SfM/main_SfMInit_ImageListing.cpp
```diff
@@ -200,6 +233,12 @@ int main(int argc, char **argv)
     }
   }
 
+  // Check if prior weights are given
+  if(cmd.used('P') && !sPriorWeights.empty())
```
pmoulon: if (

src/software/SfM/main_SfMInit_ImageListing.cpp
```diff
@@ -133,6 +165,7 @@ int main(int argc, char **argv)
       << "\t 1-> (default) view can share some camera intrinsic parameters\n"
       << "\n"
       << "[-P|--use_pose_prior] Use pose prior if GPS EXIF pose is available"
+      << "[-W|--prior_weigths] \"x;y;z;\" of weights for each dimension of the prior (default: 1.0)\n"
```
pmoulon: You said default 1.0, but we can read Vec::Zero at line 129.

src/software/SfM/main_SfMInit_ImageListing.cpp
```diff
@@ -96,6 +125,8 @@ int main(int argc, char **argv)
     sfileDatabase = "",
     sOutputDir = "",
     sKmatrix;
+  std::string sPriorWeights;
+  std::pair<bool, Vec3> prior_w_info(false, Vec3::Zero());
```
pmoulon: default must be true, (1.0,1.0,1.0) or set to this value if sPriorWeights is empty and -P is activated

pmoulon:
Merged by https://github.com/openMVG/openMVG/commit/8e3ba95af55046ac402ff1e91f8493d8276bfda3

PR #652, opened 2016-10-18
fixed bug in ApplySimilarity for control points

mitjap:
The for loop that transforms control points was accidentally nested in previous for loop causing that each control point was transformed N times (where N is number of view priors)


pmoulon:
Nice find... :+1:

PR #674, opened 2016-11-10
Added interface to pass abstract C_Progress& to some functions.

bjornpiltz:
The idea is that the caller decides how to display progress. Ideally the class C_Progress_Display should not be found in any library code, but only used in the command line tools. See main_computeMatches.
I also added a function hasBeenCanceled(), which should be checked periodically during long operations.


pmoulon:
Thank you @bjornpiltz 
I will have a look to it in the coming weeks and make a code review!

src/third_party/progress/progress.hpp
```diff
@@ -208,6 +216,11 @@ class C_Progress_display : public C_Progress
     } // display_tic
 };
 
+inline C_Progress& C_Progress::dummy()
+{
+  static C_Progress null;
```
pmoulon: Whqt do you think about (VS compile does not have guarantee to make the static a lock free operation)
```
C_Progress * instance = nullptr;
std::once_flag flag;

inline C_Progress& C_Progress::dummy()
{
  std::call_once(flag, []) { instance = new C_Progress{}; }
  return *instance;
}
```

src/third_party/progress/progress.hpp
```diff
@@ -152,45 +160,45 @@ class C_Progress
...
     std::ostream &     m_os;  // may not be present in all imps
-    /// String used to formatting display // string is more general, safer than const char *, and efficiency or size are not issues
-    const std::string  m_s1, m_s2, m_s3;
+    /// String used to formatting display
+    std::string  m_s1;
 
     /** @brief Function that check if we have to append an * in the progress bar function
     **/
-    void inc_tic()
+    void inc_tic()override
```
pmoulon: ) override
add a space

src/third_party/progress/progress.hpp
```diff
@@ -57,6 +57,10 @@ class C_Progress
       if ( !_expected_count ) _expected_count = 1;  // prevent divide by zero
     } // restart
 
+    /** @brief Indicator if the current operation should be aborted.
+     * @return Return true if the process has been canceled by the user.
+     **/
+    virtual bool hasBeenCanceled()const { return false; }
```
pmoulon: Since we introduce a new comportment it would be nice to add a unit test to check that we can stop a loop

src/third_party/progress/progress.hpp
```diff
@@ -47,7 +47,7 @@ class C_Progress
     /** @brief Initializer of the C_Progress class
      * @param expected_count The number of step of the process
```
pmoulon: help about msg must be updated

src/third_party/progress/progress.hpp
```diff
@@ -152,45 +160,45 @@ class C_Progress
...
-                                     const std::string & s1 = "\n", //leading strings
-                                     const std::string & s2 = "",
-                                     const std::string & s3 = "" )
+                                     const std::string & msg = "\n")
     // os is hint; implementation may ignore, particularly in embedded systems
-    : m_os ( os ), m_s1 ( s1 ), m_s2 ( s2 ), m_s3 ( s3 )  { restart ( ulExpected_count ); }
+    : m_os ( os ), m_s1 (msg){ restart ( ulExpected_count, msg); }
 
     /** @brief Initializer of the C_Progress_display class
      * @param expected_count The number of step of the process
```
pmoulon: help must be updated

src/third_party/progress/progress.hpp
bjornpiltz: Nice! I wasn't aware of std::call_once. I will have to add a .cpp file, right? Otherwise flag will have multiple definitions

bjornpiltz: For reference: this is only an issue with MSVC versions < 2015.  https://msdn.microsoft.com/en-au/library/hh567368.aspx#concurrencytable (see Magic Statics)

pmoulon: Ok, thx for the info.
I think it will be simpler to keep the static reference as you done it.
I was just pointing you the call_once in order to let us know that there is new interesting C++11 feature to deal with singleton allocation in multithreaded context.

src/openMVG/matching_image_collection/GeometricFilter.hpp
```diff
@@ -57,10 +58,11 @@ void ImageCollectionGeometricFilter::Robust_model_estimation
...
-  const double d_distance_ratio
+  const double d_distance_ratio,
+  C_Progress& my_progress_bar
 )
 {
-  C_Progress_display my_progress_bar( putative_matches.size() );
+  my_progress_bar.restart( putative_matches.size(), "\n- Geometric filtering -\n" );
 
 #ifdef OPENMVG_USE_OPENMP
 #pragma omp parallel for schedule(dynamic)
```
pmoulon: I think we have to deal with
```
if (my_progress_bar.hasBeenCanceled())
    continue;
```
in this loop

src/openMVG/matching_image_collection/Cascade_Hashing_Matcher_Regions_AllInMemory.cpp
```diff
@@ -144,6 +146,8 @@ void Match
 #endif
     for (int j = 0; j < (int)indexToCompare.size(); ++j)
     {
+        if (my_progress_bar.hasBeenCanceled())
+            continue;
```
pmoulon: Here you use continue, while before you used break.

src/openMVG/matching_image_collection/GeometricFilter.hpp
pmoulon: break or continue?

src/openMVG/matching_image_collection/Matcher_Regions_AllInMemory.cpp
```diff
@@ -53,6 +53,8 @@ void Matcher_Regions_AllInMemory::Match(
   for (Map_vectorT::const_iterator iter = map_Pairs.begin();
     iter != map_Pairs.end(); ++iter)
   {
+    if (my_progress_bar.hasBeenCanceled())
+        continue;
```
pmoulon: Here you use continue, while before you used break.

src/openMVG/sfm/pipelines/sfm_regions_provider.hpp
```diff
@@ -42,6 +42,12 @@ struct Regions_Provider
     for (Views::const_iterator iter = sfm_data.GetViews().begin();
       iter != sfm_data.GetViews().end() && bContinue; ++iter)
     {
+      if (my_progress_bar.hasBeenCanceled()) 
+      {
+          bContinue = false;
+          continue;
```
pmoulon: Here you use continue, while before you used break.
continue is ok, since you are using the boolean in the loop.
But since there is a multitihread context, I think the bContinue must be in a pragma critical scope

src/third_party/progress/progress.hpp
```diff
@@ -152,45 +164,46 @@ class C_Progress
...
     {
-      C_Progress::restart ( ulExpected_count ); //-- Initialize the base class
-
-      m_os << m_s1 << "0%   10   20   30   40   50   60   70   80   90   100%\n"
-      << m_s2 << "|----|----|----|----|----|----|----|----|----|----|"
-      << std::endl  // endl implies flush, which ensures display
-      << m_s3;
+      C_Progress::restart ( ulExpected_count,  msg); //-- Initialize the base class
+      if (!msg.empty())
+          m_msg = msg;
```
pmoulon: even an empty copy must work

bjornpiltz:
Regarding continue vs break: I use continue in OpenMP loops, since break is prohibited. It shouldn't be an issue  since we check at the beginning of each iteration.

src/openMVG/matching_image_collection/GeometricFilter.hpp
bjornpiltz: done

pmoulon:
My main matter about stopping the loop early (hasBeenCanceled) is that the code that follow is perhaps not robust to partial computation.
Providing the ability to have an early exit can provide some unstable behavior.
What do you think about that?

src/third_party/progress/progress.hpp
pmoulon: @bjornpiltz We must update the doxygen doc here since you added the msg variable.

src/third_party/progress/progress.hpp
pmoulon: I mean that m_msg = msg; must work even if msg is empty.
So I do not see really the need of the if clause.

src/openMVG/matching_image_collection/Cascade_Hashing_Matcher_Regions_AllInMemory.cpp
bjornpiltz: You cannot use break in an openMP loop.

src/openMVG/sfm/pipelines/sfm_regions_provider.hpp
bjornpiltz: I think it's enough to mark bContinue as volatile. It won't influence the correctness of the program anyway. Worst case: hasBeenCanceled() gets called a bunch of times.

src/third_party/progress/progress.hpp
bjornpiltz: I've added a comment that an empty string leaves the last status untouch. I don't think that an empty status makes sense semantically.

bjornpiltz:
I hope I have addressed your requests. As to "My main matter about stopping the loop early (hasBeenCanceled) is that the code that follow is perhaps not robust to partial computation."

Well, since there is no possibility to cancel the standard C_Progress_display, this would only affect those(like me) who implement their own progress/cancel class and all they have to do is check hasBeenCanceled() after processing to know if they can use the result or not.

src/openMVG/sfm/pipelines/sfm_regions_provider.hpp
```diff
@@ -88,20 +88,25 @@ struct Regions_Provider
...
+    C_Progress& my_progress_bar = C_Progress::dummy())
   {
     region_type_.reset(region_type->EmptyClone());
 
-    C_Progress_display my_progress_bar( sfm_data.GetViews().size(),
-      std::cout, "\n- Regions Loading -\n");
+    my_progress_bar.restart(sfm_data.GetViews().size(), "\n- Regions Loading -\n");
     // Read for each view the corresponding regions and store them
-    bool bContinue = true;
+    volatile bool bContinue = true;
```
pmoulon: volative vs. std::atomic<bool>?

bjornpiltz: I could go either way. The code as is should do what it's supposed to since 'volatile' guarantees that bContinue is actually evaluated each time and there is no way a bool enters an invalid state. On the other hand - the cost of one atomic access is neglible compared to reading a feature file. You decide.

zwn: `volatile` is no good for synchronization between threads on today's hardware that can do things like instruction reordering. See for example https://brooker.co.za/blog/2013/01/06/volatile.html or some threads on stack overflow discussing the differences.

pmoulon: https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rconc-volatile

pmoulon:
@bjornpiltz I think for every line
` C_Progress& my_progress_bar = C_Progress::dummy()`
we would expect to have a pointer in order to pass an interface.

I mean your PR change the default behavior. The default behavior log to the console the *** and since now you are using the  C_Progress::dummy() we loose some informative content.

Can you please udpate the PR or tell me how I can manage to recover the desired behavior?
 - force to use the C_Progress_display

pmoulon:
Please, @bjornpiltz can you squash your commits into a single one?

pmoulon:
Thank you @bjornpiltz 
The PR is now merged https://github.com/openMVG/openMVG/commit/41d1d5f5e0c983502541b86153fe47b2c0fa0feb
Sorry for the delay.

PR #688, opened 2016-11-28
[out of core]Replaced typedef PairwiseMatches with abstract class Pai…

bjornpiltz:
…rwiseMatchesContainer.

### DIFF
--- a/src/openMVG/matching/indMatch.hpp
+++ b/src/openMVG/matching/indMatch.hpp
@@ -69,10 +69,26 @@ inline std::istream& operator>>(std::istream & in, IndMatch & obj) {
 }
 
 typedef std::vector<matching::IndMatch> IndMatches;
+
+/// Pairwise matches (indexed matches for a pair <I,J>)
+/// The interface used to store corresponding point indexes per images pairs
+class PairWiseMatchesContainer
+{
+public:
+  virtual ~PairWiseMatchesContainer() {}
+  virtual void insert(std::pair<Pair, IndMatches>&& pairWiseMatches) = 0;
+};
+
 //--
 /// Pairwise matches (indexed matches for a pair <I,J>)
-/// The structure used to store corresponding point indexes per images pairs
-typedef std::map< Pair, IndMatches > PairWiseMatches;
+/// A structure used to store corresponding point indexes per images pairs
+struct PairWiseMatches : public PairWiseMatchesContainer, public std::map< Pair, IndMatches >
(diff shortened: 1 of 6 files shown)


src/openMVG/matching/indMatch.hpp
```diff
@@ -69,10 +69,26 @@ inline std::istream& operator>>(std::istream & in, IndMatch & obj) {
...
-/// The structure used to store corresponding point indexes per images pairs
-typedef std::map< Pair, IndMatches > PairWiseMatches;
+/// A structure used to store corresponding point indexes per images pairs
+struct PairWiseMatches : public PairWiseMatchesContainer, public std::map< Pair, IndMatches >
+{
+  void insert(std::pair<Pair, IndMatches> && pairWiseMatches)override
+  {
+      std::map< Pair, IndMatches >::insert(std::forward<std::pair<Pair, IndMatches>>(pairWiseMatches));
+  }
+};
```
pmoulon: You must add a serialize function for Cereal (something like this...):
```
// Serialization
  template <class Archive>
  void serialize( Archive & ar )  {
    ar(std::static_cast<std::map< Pair, IndMatches >>(*this));
  }
```

src/openMVG/matching/indMatch.hpp
```diff
@@ -85,4 +107,9 @@ inline Pair_Set getPairs(const PairWiseMatches & matches)
 }  // namespace matching
 }  // namespace openMVG
 
+namespace cereal
+{
+    // This struct specialization will tell cereal which is the right way to serialize PairWiseMatches
```
pmoulon: Does void serialize( Archive & ar )  { is still used since you have added this new specialization?

bjornpiltz:
Yes `member_serialize` tells cereal to use `serialize` as opposed to
`load`/`save`.

pmoulon:
For some other class I use serialize and don't need to add the specialization.
Does it was required for a given compiler, or does you just add it by security?

bjornpiltz:
I ran into [this compile error](https://travis-ci.org/openMVG/openMVG/jobs/179709779#L970) since load/save is defined for std::map and serialise for PairWiseMatches.

pmoulon:
Congrats your first PR was successfully merged:
https://github.com/openMVG/openMVG/commit/7d26f579beb27f3046e0c2e87ef021c3031bf7f8

Sorry for the delay.

PR #705, opened 2016-12-20
Added check if OpenCV is enabled when including KLT tracker

hcjghr:
Hi @pmoulon ,

the small corrections adds a check in the VO code for presence of OpenCV when including KLT tracker. Without this the code does not compile (at least on my computer) when OpenCV is not installed. 

Cheers,
Klemen


pmoulon:
Thx 👍

PR #719, opened 2016-12-31
Code cleanup

rperrot:
Here is a proposition of simple code cleanup : 

- Use canonical #ifdef /#define names : namespace_X_namespace_Y_file_HPP
- Force using #ifdef #define instead of #pragma once 
- Add missing comments on end of namespaces 
- Sort includes alphabetically 
- Remove the remaining using namespace directives in headers files

Hope it helps !


pmoulon:
Thx, merged :tada: :tada: :tada:

PR #741, opened 2017-01-24
Add option of UTM gps initialization

hcjghr:
Hi @pmoulon,

This is a small extension that enables to initialize priors either in ECEF or UTM coordinate system. The default option remains ECEF while UTM can be invoked using -m switch.

This comes in handy in underwater scenarios (among others) as we have different sensor estimating the depth (altitude) from lat, lon. In this way X,Y,Z are still separated and prior weights can be appropriately used. 

Best


pmoulon:
Thank you @hcjghr 
PR looks good, will merge it in the coming days

PR #743, opened 2017-01-24
[sfm] Allow passing z_near/z_far for each view to Frustum_Filter

bchretien:
Hi,

This PR may be specific to my scenario, but this should not change anything for the previously supported use cases.
In my case, I already have `zNear`/`zFar` computed for each view, and I want to use these values for the frustum filtering. Unless I'm missing something, the current code either computes these values from the structure and view observations, or initializes them all to the default values given by the user. This PR provides a third option: the  `zNear`/`zFar` map can be provided by the user, and updated if necessary (e.g. if some views are missing, the user-provided  `zNear`/`zFar` are used).

### DIFF
--- a/src/openMVG/sfm/sfm_data_filters_frustum.cpp
+++ b/src/openMVG/sfm/sfm_data_filters_frustum.cpp
@@ -23,8 +23,10 @@ using namespace openMVG::geometry::halfPlane;
 
 // Constructor
 Frustum_Filter::Frustum_Filter(const SfM_Data & sfm_data,
-  const double zNear, const double zFar)
+  const double zNear, const double zFar, const NearFarPlanesT & z_near_z_far)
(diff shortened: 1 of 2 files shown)


bchretien:
I will also make a patch to allow providing a bounding box to the filter: the objective is to keep pairs that intersect within the bounding box. Do you want me to integrate that into this PR or make a different one?

pmoulon:
Hi @bchretien The PR looks good.
Will look to merge it in the coming days.

I think it's good to do as you said, one PR per feature in order to keep thing easy to understand in the git history.

Please to another PR, so we can discuss about the design if we need to.
Moreover it allows the community to react on each PR/feature ;-)

bchretien:
Great, thanks for the feedback!

PR #745, opened 2017-01-25
[WIP][sfm] Add support for additional bounding volume filtering to Frustum_Filter

bchretien:
This PR contains the following changes:
- add a new `intersect` method that takes a vector of half-plane objects.
- add a new box-box test that shows how it can be used for multiple objects (6 boxes). Existing tests have also been updated.
- use that method to add support for bounding volumes to `Frustum_Filter::getFrustumIntersectionPairs`: it now takes an optional vector of half-plane objects used in the intersection check. For example, one can use a bounding box to only keep views whose frustums intersect within that bounding box.

This is a follow-up of #743, and will be rebased once that PR is merged (hence the `[WIP]` tag).

### DIFF
--- a/src/openMVG/geometry/frustum_box_intersection_test.cpp
+++ b/src/openMVG/geometry/frustum_box_intersection_test.cpp
@@ -38,6 +38,33 @@ TEST(box_point, intersection)
   EXPECT_FALSE( box.contains(Vec3(1,1,1)) );
 }
 
+TEST(box_box, intersection)
+{
+  double r = 1.;
+
+  // Test with a set of intersecting boxes
+  std::vector<HalfPlaneObject> boxes_ok;
+  for (int i=0; i < 6; ++i)
+  {
+    Vec3 center = Vec3::Zero();
+    center[i/2] += std::pow(-1, i%2) * r / 5.;
+    boxes_ok.emplace_back(Box(center, r));
+  }
+  EXPECT_TRUE( boxes_ok[0].intersect(boxes_ok[1]) );
+  EXPECT_TRUE( intersect(boxes_ok) );
+
+  // Test with a set of non-intersecting boxes
+  std::vector<HalfPlaneObject> boxes_ko;
+  for (int i=0; i < 6; ++i)
+  {
+    Vec3 center = Vec3::Zero();
+    center[i/2] += std::pow(-1, i%2) * 1.5 * r;
+    boxes_ko.emplace_back(Box(center, 1));
+  }
+  EXPECT_FALSE( boxes_ko[0].intersect(boxes_ko[1]) );
+  EXPECT_FALSE( intersect(boxes_ko) );
+}
+
 TEST(box_frustum, intersection)
 {
   const int focal = 1000;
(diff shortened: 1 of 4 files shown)


bchretien:
Apparently the GCC build on Travis stalled when running `openMVG_test_sfm_data_BA`. I fail to see how this is related to this PR.

src/openMVG/geometry/frustum_box_intersection_test.cpp
```diff
@@ -38,6 +38,33 @@ TEST(box_point, intersection)
   EXPECT_FALSE( box.contains(Vec3(1,1,1)) );
 }
 
+TEST(box_box, intersection)
+{
+  double r = 1.;
```
pmoulon: const double

src/openMVG/sfm/sfm_data_filters_frustum.cpp
```diff
@@ -86,9 +90,16 @@ Pair_Set Frustum_Filter::getFrustumIntersectionPairs() const
 #endif
   for (int i = 0; i < (int)viewIds.size(); ++i)
   {
+    // Prepare vector of intersecting objects (within loop to keep it
+    // thread-safe)
+    std::vector<HalfPlaneObject> objects = bounding_volume;
+    objects.insert(objects.end(),
```
pmoulon: why not consider two insert. inserting an empty object ad modifying it later on is a bit dangerous.
Else consider to user (*objects.back()) =   frustum_perView.at(viewIds[j]);
We can discuss on it.

pmoulon:
Thanks for the PR.
I like to see that you was able to even more generalize the halfplane based interestion volume.

Regarding the PR, in your case you can rebase or start from develop branch each time you work on a feature.
Travis can fail sometime, I think its due to the fact that from time to time the VM is blocked on the server side. Will try to re run the specific build manually to see how it goes.

src/openMVG/sfm/sfm_data_filters_frustum.cpp
bchretien: In the inner loop (`j`), all the elements of `objects` are constant except the last one that needs to be copied at each iteration. So the idea is to prepare `objects` with both the user-defined bounding volume and the `i`-th frustum (constant part), and reserve the last element with a dummy `HalfPlaneObject`. :+1: for using `back()` rather than the index.

src/openMVG/geometry/frustum_box_intersection_test.cpp
bchretien: :+1:

src/openMVG/sfm/sfm_data_filters_frustum.cpp
```diff
@@ -23,8 +23,10 @@ using namespace openMVG::geometry::halfPlane;
 
 // Constructor
 Frustum_Filter::Frustum_Filter(const SfM_Data & sfm_data,
-  const double zNear, const double zFar)
+  const double zNear, const double zFar, const NearFarPlanesT & z_near_z_far)
 {
+  z_near_z_far_perView = z_near_z_far;
```
pmoulon: Forgot to mention, since we are in the constructor you can use the initializer list

```
Frustum_Filter::Frustum_Filter
(
  const SfM_Data & sfm_data,
  const double zNear,
  const double zFar,
  const NearFarPlanesT & z_near_z_far
)
:  z_near_z_far_perView(z_near_z_far)
{
  ...
}
```

Issue #488, opened 2016-03-02
[build ]Compiling openMVG on NVIDIA Jetson TK1 (arm target)


polto:
Hi !

I am trying to compile openMVG on NVIDIA Jetson TK1 embedded computer and make use of it's proprietary OpenCV extensions for GPU.

cmake -DOpenMVG_USE_OCVSIFT=ON -DOpenMVG_USE_OPENCV=ON . ../openMVG/src/

Even if I use those options, openMVG try to compile also it's own SIFT implementation and it result during "make" in this error:

> Scanning dependencies of target vlsift
> [ 49%] Building C object nonFree/sift/CMakeFiles/vlsift.dir/vl/generic.c.o
> [ 49%] Building C object nonFree/sift/CMakeFiles/vlsift.dir/vl/imopv_sse2.c.o
> /media/ubuntu/ed6746e6-6af4-40f4-80cf-51624865dfef/code/openMVG/src/nonFree/sift/vl/imopv_sse2.c:15:2: error: #error "Compiling with SSE2 enabled, but no __SSE2__ defined"
>  #error "Compiling with SSE2 enabled, but no __SSE2__ defined"
>   ^
> /media/ubuntu/ed6746e6-6af4-40f4-80cf-51624865dfef/code/openMVG/src/nonFree/sift/vl/imopv_sse2.c:22:23: fatal error: emmintrin.h: No such file or directory
>  #include <emmintrin.h>
>                        ^
> compilation terminated.
> make[2]: *** [nonFree/sift/CMakeFiles/vlsift.dir/vl/imopv_sse2.c.o] Error 1
> make[1]: *** [nonFree/sift/CMakeFiles/vlsift.dir/all] Error 2
> make[1]: *** Waiting for unfinished jobs....
> Linking CXX static library ../../../Linux-armv7l-Release/libopenMVG_kvld.a
> [ 49%] Built target openMVG_kvld
> Linking CXX static library ../../Linux-armv7l-Release/libopenMVG_multiview_test_data.a
> [ 49%] Built target openMVG_multiview_test_data
> Linking CXX static library ../../Linux-armv7l-Release/libopenMVG_matching.a
> [ 49%] Built target openMVG_matching
> make: *** [all] Error 2

I do not have SSE2, I have a GPU and I want to use GPU optimised SIFT implementation in openCV.

I just tried to comment the nonFree stuff. but "-lvlsift" strike back later ..

[...]

PR #752, opened 2017-02-02
[build] fix build for ARM

jkerfs:
There were two issues that occurred while compiling on the Jetson TX1 (64-bit ARM architecture):

1. undefined reference to `_vl_cpuid`.  This was originally mentioned in https://github.com/openMVG/openMVG/issues/488 and fixed in https://github.com/openMVG/openMVG/commit/cbc35f63293ea053a85916a45441143f2727d011, but then the issue still persists.  I re-added this solution to the `host.c`.  I also submitted https://github.com/vlfeat/vlfeat/pull/150 to address the issue in vlfeat.

2. CMake attempts to use SSE even when the architecture does not support it.  I fixed this in `CMakeLists.txt` by removing the OR clause.  Is there any reason why SSE should be used when `SSE2_FOUND` is not defined?

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -91,7 +91,8 @@ set(CMAKE_MODULE_PATH
   ${CMAKE_MODULE_PATH} ${CMAKE_CURRENT_SOURCE_DIR}/cmakeFindModules)
 include(OptimizeForArchitecture)
 OptimizeForArchitecture()
-if (SSE2_FOUND OR TARGET_ARCHITECTURE STREQUAL "native")
+
+if (SSE2_FOUND)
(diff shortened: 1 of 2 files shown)


pmoulon:
Thanks.
This change will be merge to develop.
FYI, I push only release version to master.

PS: I wonder how works OpenMVG's Akaze and Sift implementation compare to Vlfeat Sift on ARM.

pmoulon:
Thank you @jkerfs 
I merged the PR in the develop_matching branch https://github.com/openMVG/openMVG/commit/c4ad0eb0348d07c0851f604be20baa2b2dc0bc51 and I will merge it to develop soon.
It will be included in master when I will create the next release candidate v1.2.

PR #755, opened 2017-02-03
Avoid API clash between clang and gcc (fix #753)

### LINKED ISSUE #753: Issue with OPENMVG_STD_UNORDERED_MAP
In `src/openMVG/types.hpp`, `OPENMVG_STD_UNORDERED_MAP` is defined for clang, thus changing the actual type used for `Hash_Map` (`std::unordered_map` instead of `std::map`). As a result, if one compiles openMVG with gcc, installs it, and later compiles a program that depends on openMVG with clang, linking will fail since some symbols will not be found (methods with `Hash_Map` parameters for instance).

If that compile-time branching is unavoidable (a `std::map` Cereal serialization bug was mentioned in the commits), we should set this definition in CMake, and export it to the CMake configuration files. That way, programs depending on openMVG will automatically have `-DOPENMVG_STD_UNORDERED_MAP` when relevant. `types.hpp` would need to be adapted (`#ifdef OPENMVG_STD_UNORDERED_MAP` rather than `#ifdef __clang__`).

bchretien:
- `OPENMVG_STD_UNORDERED_MAP` is now defined by CMake when clang is detected.
- The compile definitions are forwarded to `openMVG-targets.cmake` and `openMVGConfig.cmake`, thus required `CFLAGS` are used automatically by projects depending on openMVG (set as compile definitions for each openMVG target, and summarized in a new CMake variable `OpenMVG_CFLAGS`). A new macro was added to this end, `register_definitions`, which calls `add_definitions` and stores the definition for later use. I only replaced `add_definitions` calls when it seemed relevant and safe (i.e. when forwarding them seemed important). I'm hesitating to add `OPENMVG_USE_OPENMP` that way (forces OpenMP use down the line, possibly breaking existing builds).

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -76,6 +76,21 @@ elseif()
   set(BUILD_SHARED_LIBS OFF)
 endif()
 
+# Macro used to add definitions, and automatically forward them to targets
+# imported with OpenMVGConfig.cmake.
+set(OpenMVG_DEFINITIONS "")
+MACRO(register_definitions DEF)
+  add_definitions(${DEF})
+  string(REPLACE "-D" "" STRIPPED_DEF ${DEF})
+  list(APPEND OpenMVG_DEFINITIONS ${STRIPPED_DEF})
+ENDMACRO()
+
+# For both regular Clang and AppleClang
+if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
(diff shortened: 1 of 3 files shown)


pmoulon:
Thank you @bchretien 
I merged the PR here https://github.com/openMVG/openMVG/commit/a7c8f57d729127eaea402a0b930801dc3e015a22 and will merge it to develop branch soon

PR #757, opened 2017-02-04
Geometry IO ply

rperrot:
Hi, 

this PR aims at creating a new sublibrary geometry::io with the support of saving and loading various point cloud formats. This PR only concerns ply file type (as it is already used in openMVG). 

It supports : reading and writing ply files either in ascii or binary format (little and big endian are supported). It supports point cloud with only points, points with normals, points with colors and points with normals and colors. 

There are some limitations for now : 
- Data must be in canonical order : (ie we must have ptX ptY ptZ [normX normY normZ] [R G B], element could not be interleaved)
- Color data is only supported using uchar data 
- Point and normal data are only supported in float or double format (no integer support)

However it should cover most of openMVG usages. 

The library aims also to be simple enough for the users : just use PLY_Read or PLY_Write in ply.hpp file. 

Hope it helps !

### DIFF
--- a/src/openMVG/geometry/io/ply.cpp
+++ b/src/openMVG/geometry/io/ply.cpp
@@ -1,0 +1,430 @@
+#include "ply.hpp"
+#include "ply_helper.hpp"
+
+#include <algorithm>
+#include <cctype>
+#include <fstream>
+#include <functional>
+#include <locale>
+
+namespace openMVG
+{
+namespace geometry
+{
+  namespace io
+  {
+    /**
+    * @brief Export a set of points to a PLY file 
+    * @param pts A list of points 
+    * @param nor A list of normal (one for each points) 
+    * @param col A list of color (one for each points) 
+    * @param path Path of the file to save 
+    * @param binary Indicate if the file should be written to binary format 
+    * @retval true if load is successful 
+    * @retval false if load fails 
+    */
+    bool PLYWrite( const std::vector<Vec3> &pts,
+                   const std::vector<Vec3> *nor,
+                   const std::vector<Vec3uc> *col,
+                   const std::string &path,
+                   const bool binary )
+    {
+      std::ofstream file( path );
+      if ( !file )
+      {
+        return false;
+      }
+
+      // Write header
+      file << "ply" << std::endl;
+      if ( binary )
+      {
+        file << "format binary_little_endian 1.0" << std::endl;
+      }
+      else
+      {
+        file << "format ascii 1.0" << std::endl;
+      }
+
+      const size_t nb_vert = pts.size();
+      file << "element vertex " << nb_vert << std::endl;
+      file << "property double x" << std::endl;
+      file << "property double y" << std::endl;
+      file << "property double z" << std::endl;
+      if ( nor )
+      {
+        file << "property double nx" << std::endl;
+        file << "property double ny" << std::endl;
+        file << "property double nz" << std::endl;
+      }
+      if ( col )
+      {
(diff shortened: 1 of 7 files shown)


src/openMVG/numeric/eigen_alias_definition.hpp
```diff
@@ -70,6 +70,9 @@ namespace openMVG
   /// 3d vector using double internal format
   using Vec3 = Eigen::Vector3d;
 
+  /// 3d vector using unsigned char internal format 
+  using Vec3uc = Eigen::Matrix<unsigned char,3,1> ;
```
pmoulon: be careful to add its aligned constraint for vector below:
```
EIGEN_DEFINE_STL_VECTOR_SPECIALIZATION_INITIALIZER_LIST(Vec3uc)
```

src/openMVG/geometry/io/ply.cpp
```diff
@@ -0,0 +1,700 @@
...
+    * @retval true if load is successful 
+    * @retval false if load fails 
+    */
+    bool PLYWrite( const std::vector<Vec3> &pts,
+                   const std::vector<Vec3> *nor,
+                   const std::vector<Vec3uc> *col,
+                   const std::string &path,
+                   const bool binary )
+    {
+      std::ofstream file( path );
```
pmoulon: Do you need to open the file in binary mode (since you can write in binary into it)?

rperrot: I don't know exactly, because it's mixed : text and binary. I think the filesystem manage it correcly depending on the underlying function (operator<<() or write()). We may need to check what the standard say about that.

jkerfs:
Is there a reason not to use [PointCloudLibrary](https://github.com/PointCloudLibrary/pcl) for this?  I have been using openMVG and PCL for reading/writing PLY files with good success.  

It is fairly straightforward to convert a PointCloud to other formats: http://pointclouds.org/documentation/tutorials/basic_structures.php#basic-structures.

Parsing and formatting PLY files can be error-prone, so it might not make sense to duplicate the functionality of PCL here.  Although it is always nice to reduce the number of 3rd party dependencies.

pmoulon:
@jkerfs 
The goal is to minimize the external dependencies. PLY is quite standard and so it's valuable to have our own interface to save/load some point clouds.
Moreover it allow to load some data for ongoing development without relying on external dependencies.

rperrot:
@jkerfs You've just hightlight the reason : reducing the number of 3rd party libraries. I don't have anything against PCL, but I don't know how it compiles on various systems (from ARM to i64, using linux, wndows, osx, ios). Moreover it seems that it depends on other libraries that openMVG still not requires (boost for example, that is known to be difficult to compile for begginers on Windows). 

Another reason is to provide additional information that could not be added simply (the final goal is to provide more than ply), for example using lidar data you can have focal info, precision, etc. I don't know if PCL handles it.

Final reason is that PCL is (by itself) a huge library (I've just dowloaded it, it's 143Mo to add to the repository, maybe too big for just reading ply files).

jkerfs:
@rperrot Thank you for the quick breakdown of reasons!  Those are all valid points.

I can tell you from experience that compiling PCL on ARM is a several hour ordeal that is best avoided!  Therefore, I think you are right that this will be a valuable addition since it dramatically reduces the complexity associated with PCL.  Users who need the power of PCL can always link with openMVG on their own system.

As a side note, this is nice implementation - considerably more straightforward than the analogous code in PCL.

PR #763, opened 2017-02-05
Add DJI Mavic Pro to sensor_width_database

spikeheap:
From https://www.dji.com/mavic/info, the Mavic Pro has a 1/2.3” (CMOS), which converts to 6.17mm sensor width.


pmoulon:
Thank you.
We must add the two version in the database, since some users reported that depending of the firmware the EXIF camera was not always the same.
So please add:
```
DJI FC220;6.17
FC220;6.17
```
as we have done for the other model (thanks)
```
 DJI FC300S;6.16
 DJI FC300X;6.16
 DJI FC350;6.17
 DJI PHANTOM VISION FC200;6.17
 FC300S;6.16
 FC300X;6.16
 FC350;6.17
```

spikeheap:
Hi @pmoulon, thanks – I'm not sure how I didn't notice that pattern the first time round.

I've added both variants into this PR now, and tested that `FC220;6.17` works for the current firmware release for the Mavic Pro. 

Cheers!

Issue #750, opened 2017-01-30
Add OpenCV version requirement to CMakeLists.txt


jkerfs:
Since the AKAZE detector require OpenCV 3+ (see https://github.com/openMVG/openMVG/issues/561), does it make sense to add the OpenCV version requirement to `src/CMakeLists.txt`?  Otherwise the error messages are somewhat cryptic, and it is not obvious that the OpenCV version is incorrect.

Here are two possible ways to do it:
<pre><code>
...
if (OpenMVG_USE_OPENCV)
  FIND_PACKAGE( OpenCV <b>3.0.0 QUIET</b> )
  if (NOT OpenCV_FOUND)
    message(STATUS "OpenCV was not found. Disabling OpenCV support.")
...
</pre></code>

<pre><code>
...
if (OpenMVG_USE_OPENCV)
  FIND_PACKAGE( OpenCV)
  if (NOT OpenCV_FOUND OR <b>OpenCV_VERSION VERSION_LESS "3.0.0"</b>)
    message(STATUS "OpenCV was not found. Disabling OpenCV support.")
...
</pre></code>

pmoulon:
Hi @jkerfs 
Thank you for the report and the suggestions.
You're right the warning about the version could be a good idea.

PS: Please note also that OpenMVG have its own AKAZE version.

PR #775, opened 2017-02-08
check OpenCV version in CMakelists.txt

jkerfs:
Simple resolution of https://github.com/openMVG/openMVG/issues/750

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -361,7 +361,7 @@ endif()
 # - only external and enabled only if OpenMVG_USE_OPENCV is set to ON
 # ==============================================================================
 if (OpenMVG_USE_OPENCV)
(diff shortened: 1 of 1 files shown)


pmoulon:
Does it is possible to check that at least opencv version > 3.0 is present.
Actually OpenCV release is 3.2 and not 3.0 (VERSION_GREATER?)

jkerfs:
The default behavior of `find_package(OpenCV [VERSION] QUIET)` is to accept any version of OpenCV that is compatible with `[VERSION]`.  I am running OpenCV 3.2 on my computer and this check passes just fine for me.

https://cmake.org/cmake/help/v3.0/command/find_package.html

pmoulon:
Ok, thx.

PR #776, opened 2017-02-09
Add snapcraft.yaml file

mardy:
This is the recipe for creating a snap package of openMVG.
See https://snapcraft.io for more information.


pmoulon:
Thank you for this contribution.
It's nice to have it, I would like to request you to consider also install the sensor_width_camera_database.txt file.
Here the install rule used in the CMake configuration
https://github.com/openMVG/openMVG/blob/master/src/openMVG/exif/sensor_width_database/CMakeLists.txt#L8
Adding a single copy instruction must do the trick.

mardy:
On 10/02/2017 08:37, Pierre Moulon wrote:
> It's nice to have it, I would like to request you to consider also
> install the sensor_width_camera_database.txt file.

Oh, indeed! I'll add that tonight!

mardy:
I've updated the branch and added the camera DB file to the package.

pmoulon:
Thank you very much, will check it and merge it to develop for the next release.

mardy:
Hi Pierre! Just in case you are wondering how to test the package: make sure you have the `snap` command available; if not, you can follow [these instructions](https://snapcraft.io/docs/core/install).
Once you have `snap` installed, you can just run
```
snap install openmvg
```
And the package will be installed. Note that in order to avoid namespace issues, all command are prefixed with `openmvg`; so, for example, in order to run `GlobalSfM` you'll actually need to type
```
openmvg.GlobalSfM
```
On the positive side, one can simply type `openmvg.` and then press TAB twice in the terminal to see the whole list of available OpenMVG tools.

rjanvier:
@mardy the cmake script detect cpu instructions (SSE, AVX etc) how this his handled by canonical buildbots?

mardy:
On 12/02/2017 19:21, Romain Janvier wrote:
> @mardy <https://github.com/mardy> the cmake script detect cpu
> instructions (SSE, AVX etc) how this his handled by canonical buildbots?

Are you sure about it? I've just rebuilt it to double-check, and I
didn't see any mention of checking for SSE by cmake.

In the general case, I wish the detection could actually happen at
runtime; if that's not possible, then I guess we could publish the snap
under different names (openMVG, openMVG-sse, openMVG-avx, etc.)
depending on the situation.

rjanvier:
https://github.com/openMVG/openMVG/blob/master/src/cmakeFindModules/FindSSE.cmake
 I can't check for now but maybe @pmoulon can confirm you.

pmoulon:
Here the line that I call
https://github.com/openMVG/openMVG/blob/master/src/CMakeLists.txt#L93

Since I do not call AutodetectHostArchitecture, I don't think that fancy CPU support is enable by now.

Default setting must just enable native build and SSE extension if the CPU is capable (SSE is pretty common now on every computer).

Since the project is working fine on ARM target too, I think the cmake detection is working right.

@mardy I think the best would be to put the snap file into a dir called ./dist and certainly that a tiny readme.md would be nice in order to explain to people how to create/obtain/How to use the snap.

rjanvier:
I agree @pmoulon, but since snaps are build on a buildfarm (for users that don't build them themself)  this may lead to non reproductible builds. Worst, binary could be unusable and crash on host machine that do not support these instructions. Since you only use SSE2 it should be ok everywhere (more specially on 64 bits) now but I don't know how it's handled at third parties level (Eigen, Ceres, vlfeat...)

mardy:
Hi! I moved the file as requested, added a README, and updated the recipe to include a couple of useful python scripts.

mardy:
It seems the binary I've built indeed doesn't work on all machines (I've built it on an AMD CPU, and now I see that all programs crash at start on an Intel i7 ("illegall instruction").
This is the 'flags' line from my `/proc/cpuinfo`:
```
flags		: fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx rdtscp lm constant_tsc arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc aperfmperf pni dtes64 monitor ds_cpl vmx smx est tm2 ssse3 cx16 xtpr pdcm sse4_1 sse4_2 popcnt lahf_lm tpr_shadow vnmi flexpriority ept vpid dtherm ida
```
Do I need to explicitly disable AVX when building? Could that be the cause of the error?

pmoulon:
Adding ```-DTARGET_ARCHITECTURE=generic``` to the cmake command line should do the trick.
Can you try it?

mardy:
Hi Pierre! Thanks, that option helps :-)
I pushed a new commit which adds that option. Out of curiosity, how hard would it be to detect AVX at runtime?

pmoulon:
Detecting AVX at runtime is not too hard.
You can do something similar to that https://github.com/openMVG/openMVG/commit/64474312ec075075eaf5cb919d91eae4f0e45512 (here its sse4.2 check)

See https://insufficientlycomplicated.wordpress.com/2011/11/07/detecting-intel-advanced-vector-extensions-avx-in-visual-studio/

pmoulon:
Thanks,
Merged by https://github.com/openMVG/openMVG/commit/49e7a163ce7196feb2e410dc0534e15bee2721b0

PR #793, opened 2017-02-20
Use the CMAKE_CURRENT_SOURCE_DIR for the config

8W9aG:
* If you include openMVG from another project and
add it as a subdirectory using cmake you will notice
it tries to get the OpenMVGConfig.cmake.in from
the root of your project rather than the root of
the openMVG project
* This ensures that it gets the OpenMVGConfig.cmake.in
from the root of the openMVG project

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -618,7 +618,8 @@ endif (OpenMVG_USE_INTERNAL_CERES)
 # Create a OpenMVGConfig.cmake file. <name>Config.cmake files are searched by
 # FIND_PACKAGE() automatically. We configure that file so that we can put any
 # information we want in it, e.g. version numbers, include directories, etc.
-configure_file("${CMAKE_SOURCE_DIR}/cmakeFindModules/OpenMVGConfig.cmake.in"
+set(OPENMVG_ROOT_DIR ${CMAKE_CURRENT_SOURCE_DIR})
+configure_file("${OPENMVG_ROOT_DIR}/cmakeFindModules/OpenMVGConfig.cmake.in"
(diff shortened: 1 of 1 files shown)


rperrot:
Maybe only switching from CMAKE_SOURCE_DIR to CMAKE_CURRENT_SOURCE_DIR fix the issue ? 

note : @pmoulon we must look in all the source to replace the CMAKE_SOURCE_DIR with CMAKE_CURRENT_SOURCE_DIR (same for CMAKE_BINARY_DIR -> CMAKE_CURRENT_BINARY_DIR)

pmoulon:
Thank you for your PR.

When OpenMVG is used a third party library, I think it's better to rely on a local of system install rather than include it as a cmake subdirectory. https://github.com/openMVG/openMVG/blob/master/BUILD#L127

Perhaps I did not catch exactly how you are using it.

8W9aG:
I was using add_subdirectory, I will just use find_package instead

rperrot:
@8W9aG even if you close this PR, I'd like to thank you for your suggestion because it highlights some importants cmake usages. 

Moreover, even if you're using openMVG with find_package, I personally don't see any reason why we forbid the usage of add_subdirectory. (but it's a personally POV).

pmoulon:
I agree with @rperrot You can keep the PR open and check if your change is still working in the case of using OpenMVG as a third_party module.
We are working hard on trying to simplify the cmake usage (https://github.com/openMVG/openMVG/issues/798) so it's the good time to test your option and see if it works in all the case ;-)

8W9aG:
Reopened

PR #818, opened 2017-03-10
Add possibility to initialize GPS from file

hcjghr:
Hi @pmoulon,

in this PR I added the possibility of initializing GPS priors with the file instead of reading it from EXIF tags. The file has to have a header and for each image a line in the format of:

[img_name;lat;lon;alt;]
or
[img_name;utm_x;utm_y;alt;]

example:

image_name; utm_x; utm_y; alt;
cap02_00003.jpg; 564206.1144497304; 4127162.4705770193; -1695.2418532106;
cap02_00004.jpg; 564206.4317824967; 4127162.6070583109; -1695.3906190200;

Header is currently not used, but I assume the file has it (for clarity reasons) but could be used to automatically detect the format (GPS/UTM) of the input. Currently it is detected through -m (i_GPS_XYZ_method) switch as before.

The priors given in the file have priority over the EXIF tag as this enables the user to easily override what is in the exif tag. Every image which does not have information in the text file will still look into an exif tag if it has a valid gps prior.

I find this useful when extraction gps data from another source as it removes the need to embed the gps tag in exif for each image.

Hope you find it useful.

Klemen


pmoulon:
Thank you @hcjghr.

We could think to add a third mode to directly send X,Y,Z data.

src/software/SfM/main_SfMInit_ImageListing.cpp
```diff
@@ -491,7 +492,8 @@ int main(int argc, char **argv)
     }
 
     // Build the view corresponding to the image
-    const std::pair<bool, Vec3> gps_info = ( cmd.used('G') && !map_GPS_data.empty() ) ? checkGPS(*iter_image, map_GPS_data, i_GPS_XYZ_method) : checkGPS(sImageFilename, i_GPS_XYZ_method);
+    //const std::pair<bool, Vec3> gps_info = ( cmd.used('G') && !map_GPS_data.empty() ) ? checkGPS(*iter_image, map_GPS_data, i_GPS_XYZ_method) : checkGPS(sImageFilename, i_GPS_XYZ_method);
+    const std::pair<bool, Vec3> gps_info = std::make_pair<bool, Vec3>(true,Vec3(1,1,1));
```
pmoulon: `const std::pair<bool, Vec3> gps_info(true, Vec3::Constant(1));`

src/software/SfM/main_SfMInit_ImageListing.cpp
```diff
@@ -491,7 +492,8 @@ int main(int argc, char **argv)
     }
 
     // Build the view corresponding to the image
-    const std::pair<bool, Vec3> gps_info = ( cmd.used('G') && !map_GPS_data.empty() ) ? checkGPS(*iter_image, map_GPS_data, i_GPS_XYZ_method) : checkGPS(sImageFilename, i_GPS_XYZ_method);
+    //const std::pair<bool, Vec3> gps_info = ( cmd.used('G') && !map_GPS_data.empty() ) ? checkGPS(*iter_image, map_GPS_data, i_GPS_XYZ_method) : checkGPS(sImageFilename, i_GPS_XYZ_method);
```
pmoulon: Do we need to keep the commented line?

hcjghr: No. The only one should be: 
const std::pair<bool, Vec3> gps_info = ( cmd.used('G') && !map_GPS_data.empty() ) ? checkGPS(*iter_image, map_GPS_data, i_GPS_XYZ_method) : checkGPS(sImageFilename, i_GPS_XYZ_method);

Line  const std::pair<bool, Vec3> gps_info = std::make_pair<bool, Vec3>(true,Vec3(1,1,1)); was just for testing and I overlooked that I didn't deleted it out before.

ssampang:
@hcjghr, thanks for adding this capability. Could you tell me whether you noticed faster processing times when using GPS priors? As I mentioned in #1080, I'm noticing slower runtimes when running `openMVG_main_IncrementalSfM` with GPS priors, and I'm wondering if that's (potentially) normal.

PR #839, opened 2017-03-31
Parallelize MVE2 export

seepa:
With this code change, the MVE2 exporter gets refactored to first write the complete `synth_0.out` file and afterwards export all views in parallel.

### DIFF
--- a/src/software/SfM/main_openMVG2MVE2.cpp
+++ b/src/software/SfM/main_openMVG2MVE2.cpp
@@ -26,6 +26,7 @@ using namespace openMVG::features;
 #include <cmath>
 #include <iterator>
 #include <iomanip>
+#include <atomic>
 
 /// Naive image bilinear resampling of an image for thumbnail generation
 template <typename ImageT>
@@ -41,7 +42,7 @@ create_thumbnail
  * - An MVE2 scene appears to duplicate camera rot matrix and trans vector per-view data in 'meta.ini'
  *   within the first section of 'synth_0.out'.
  * - We do not save the original, instead we rely on the undistorted image from openMVG.
- * - We do not output thumbnails or EXIF blobs, as these appear only to be used only for the GUI UMVE.
+ * - We do not output EXIF blobs, as these appear only to be used for the GUI UMVE.
  * - To avoid encoding loss, openMVG images should be written as .PNG if undistorted images are *not* computed.
  * - In OpenMVG, some views may have some missing poses; MVE does *not* require a contiguous camera index.
  *
@@ -54,7 +55,7 @@ bool exportToMVE2Format(
   const std::string & sOutDirectory // Output MVE2 files directory
   )
 {
-  bool bOk = true;
+  std::atomic<bool> bOk(true);
   // Create basis directory structure
   if (!stlplus::is_folder(sOutDirectory))
   {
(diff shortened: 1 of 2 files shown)


donlk:
I was just about to do this today. Guess not anymore, thanks!

pmoulon:
Thank you @seepa Will merge it in the develop branch for the future release.
Fun fact, I done also the change on progress.hpp some days ago but did not pushed it yet ;-)

src/software/SfM/main_openMVG2MVE2.cpp
```diff
@@ -71,181 +71,201 @@ bool exportToMVE2Format(
...
     out << "drews 1.0\n";  // MVE expects this header
     out << cameraCount << " " << featureCount << "\n";
 
-    // Export (calibrated) views as undistorted images
-    C_Progress_display my_progress_bar(sfm_data.GetViews().size());
-    Image<RGBColor> image, image_ud, thumbnail;
-    std::string sOutViewIteratorDirectory;
-    for(Views::const_iterator iter = sfm_data.GetViews().begin();
-      iter != sfm_data.GetViews().end(); ++iter, ++my_progress_bar)
+    for(Views::const_iterator iter = views.begin(); iter != views.end(); ++iter)
```
pmoulon: You can use c++11 range loop
```
for (const auto & view : views)
```

src/software/SfM/main_openMVG2MVE2.cpp
```diff
@@ -71,181 +71,201 @@ bool exportToMVE2Format(
...
+        stlplus::folder_create(sOutViewsDirectory);
+    }
+
+    C_Progress_display my_progress_bar(views.size());
+
+    #pragma omp parallel for schedule(dynamic)
+    for (std::size_t i = 0; i < views.size(); ++i)
     {
-      const View * view = iter->second.get();
+      if (!bOk) continue;
```
pmoulon: If we hit the case of continue we will have incorrect progress bar behavior.

src/software/SfM/main_openMVG2MVE2.cpp
```diff
@@ -71,181 +71,201 @@ bool exportToMVE2Format(
...
+    if (!stlplus::folder_exists(sOutViewsDirectory))
+    {
+        cout << "\033[1;31mCreating directory:  " << sOutViewsDirectory << "\033[0m\n";
+        stlplus::folder_create(sOutViewsDirectory);
+    }
+
+    C_Progress_display my_progress_bar(views.size());
+
+    #pragma omp parallel for schedule(dynamic)
+    for (std::size_t i = 0; i < views.size(); ++i)
```
pmoulon: omp loop on VisualStudio compiler does not like unsigned index, So it's better to use an int for i and cast the `views.size()` to an int with a static_cast<int>

src/software/SfM/main_openMVG2MVE2.cpp
seepa: If there is an error while exporting the views, all currently running threads will finish their work (and update the progress at the end of the loop). For all remaining threads, this statement ensures they exit without doing any work. This behavior is akin to the previous, serial version of the for-loop, where upon failure the loop would end with a `return EXIT_FAILURE`.

seepa: Should we instead continue exporting the rest of the views, even if there was an error along the way?

pmoulon: If we hit continue, we go to the next thread and so does not hit the progress bar ++ function. With this bOk variable I tried to implement an early exit, and you are right I think it would be better to stop and return EXIT_FAILURE since we met a problem during the export. I think also that it would be better to have this value as an atomic<bool> ;-)

Can you update the PR in order to change the bool bOk to an atomic and handle the EXIT_FAILURE or simply remove the bOK variable and directly return EXIT_FAILURE the first time we hit an error?

seepa: I can change `bOK` to an atomic. As far as I know you can't do a `return EXIT_FAILURE` in an openmp for loop, that's why I went for the current approach in the first place. The `continue` at the beginning serves as an early exit for all remaining threads. Not updating the progress in that case seems logical, as no progress is made anyways...

pmoulon: LGTM will merge your PR during the week 😄

pmoulon:
Thank you @seepa: The PR is merged in develop and will be pushed in the next release ;-)
Hoping to see future contribution :laughing: 

@donlk Ready to test ;-)

src/software/SfM/main_openMVG2MVE2.cpp
```diff
@@ -71,181 +72,201 @@ bool exportToMVE2Format(
...
+    }
+
+    C_Progress_display my_progress_bar(views.size());
+
+    #pragma omp parallel for schedule(dynamic)
+    for (int i = 0; i < static_cast<int>(views.size()); ++i)
     {
-      const View * view = iter->second.get();
+      if (!bOk) continue;
+      const View * view = views.at(i).get();
```
pmoulon: @seepa:

Note that if there is some non contiguous view this code can crash!
I mean views index can be sometimes not 0,1,2 but something like 10,20,30 (if we use only a subpart of an existing sfm_data)

I think that something like this would be better:
```
for (int i = 0; i < static_cast<int>(views.size()); ++i)
{
    auto view_it = views.begin();
    std::advance(view_it, i);
    const View * view = view_it->second.get();
```

seepa: You're right. I overlooked the fact that views is a hashmap. I'll update the pull request...

PR #843, opened 2017-04-05
Fix view iteration in MVE2 export

seepa:


pmoulon:
Thanks @seepa

pmoulon:
I need to apply similar things for every exporter ;-)

PR #847, opened 2017-04-06
Define NCOMPLEX also for users of cxsparse.

zwn:
Make NCOMPLEX defined not only when compiling cxsparse but also when
using it. Fixes some warnings and makes compiling and using cxsparse
consistent.

### DIFF
--- a/src/third_party/cxsparse/CMakeLists.txt
+++ b/src/third_party/cxsparse/CMakeLists.txt
@@ -10,9 +10,10 @@ INCLUDE_DIRECTORIES(./
 
 FILE(GLOB CXSparse_SRCS "Source/*.c")
 set_source_files_properties(${CXSparse_SRCS} PROPERTIES LANGUAGE C)
-add_definitions(-DNCOMPLEX)
 
 ADD_LIBRARY(cxsparse STATIC ${CXSparse_SRCS})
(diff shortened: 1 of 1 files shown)


src/third_party/cxsparse/CMakeLists.txt
```diff
@@ -10,9 +10,10 @@ INCLUDE_DIRECTORIES(./
 
 FILE(GLOB CXSparse_SRCS "Source/*.c")
 set_source_files_properties(${CXSparse_SRCS} PROPERTIES LANGUAGE C)
-add_definitions(-DNCOMPLEX)
 
 ADD_LIBRARY(cxsparse STATIC ${CXSparse_SRCS})
+target_compile_definitions(cxsparse PUBLIC NCOMPLEX)
```
pmoulon: Do we need PRIVATE or PUBLIC?

zwn: According to my understanding PUBLIC applies the defines for both uses - internally and for dependents. PRIVATE should apply just internally and INTERFACE just for dependents. We want NCOMPLEX also defined when header from cxsparse is included by some other code - thus PUBLIC. 

But take that with a grain of salt, since this is one of mine first endeavors into cmake. I am basing that on https://cmake.org/cmake/help/v3.0/manual/cmake-buildsystem.7.html#target-usage-requirements.

PR #848, opened 2017-04-06
Disable warning about size_t conversions for ceres

zwn:
There are many warnings C4267 when compiling with msvc2015 64bit about
conversion from size_t to smaller data types, possible loss of data.
There is not much we can do about it, so we just disable the warning
here.


zwn:
Actually even better solution seems to be just putting

```
if (MSVC)
  # Disable warning about size_t conversions for all third_party libraries.
  add_compile_options(/wd4267)
endif()
```

into openMVG/src/third_party. This way the setting is inherited by all targets defined there (there are more offenders in the third_party directory).

Would it make sense to disable all warnings for third_party? It would be as easy as replacing `/wd4267` with `/w`. I guess there is not much the user of openMVG can do about such warnings.

pmoulon:
Hi @zwn 
I will close this one since this warning is already handle by 
https://github.com/openMVG/openMVG/pull/852

We will reopen this one if we see that we need it.

Issue #844, opened 2017-04-05
warning C4267: 'argument': conversion from 'size_t'


zwn:
I am compiling using win7+msvc2015 64bit. With this setup `size_t` is 64bit number but `IndexT` is `uint32_t`. This causes a lot of warning about possible loss of data when `size_t` is assigned into `IndexT`. The responsible warning is [C4267](https://msdn.microsoft.com/en-us/library/6kck0s93.aspx)

Are you aware of the issue? What is the best strategy to deal with it? I could find 4 places where the warning is turned of using a pragma directive. There is so many such messages emitted that it makes the compiler output almost useless.

I see 3 possible ways to go:
 - One way to go could be disabling the warning from cmake for the whole project but it somehow does not feel right since the assignment really does drop data and it would be turned off for everything - not just the `IndexT` type where it might be ok.
 - So another way could be making the IndexT into a proper type and add some operator overloads with explicit type cast inside but that might be a lot of work and some other problems might be uncovered along the way. 
 - Making the IndexT be 64bit value would probably unnecessarily increase memory usage (?) but it would be really simple and the result would be correct.

pmoulon:
Hi @zwn 
I'm aware of this issue.

IndexT should remains a `uint32_t`. If you look to the branch called develop, I think there is already less warnings than in master.
The proper way would be to check the type everywhere and do the appropriate casting everywhere (static_cast<>) but as you said it can introduce some side effect (unit test can help to see if we made something wrong).

BTW, notice that many of those C4267 are from the ceres-solver library.

So:
1. Can you take a look at the develop branch and tell us if there is less warning than before?
Then we can decide which way to go (global pragma or careful handling of every warning).

Thanks.

zwn:
Actually, I was compiling branch develop already, since I didn't want to report something that is possibly fixed. Sorry for not mentioning it right away. I cannot tell if there is less warnings since I don't have a base line.

I have also noticed that many of the warnings come from the third party libraries.

pmoulon:
What about using the cmake pragma things only for the third party libraries and see how many left there is for OpenMVG core ;-)?

zwn:
Actually, part of this is already there, see 

https://github.com/openMVG/openMVG/blob/master/src/third_party/ceres-solver/CMakeLists.txt#L556

where it says:

```
  # Disable warning about int64 to int32 conversion. Disabling
  # this warning may not be correct; needs investigation.
  # TODO(keir): Investigate these warnings in more detail.
[...]

PR #852, opened 2017-04-07
Disable warning about size_t conversions for 3rd party libraries

zwn:
There are many warnings C4267 when compiling with msvc2015 64bit about
conversion from size_t to smaller data types, possible loss of data.
There is not much we can do about it for third_party libraries here, so
we just disable the warning.

### DIFF
--- a/src/third_party/CMakeLists.txt
+++ b/src/third_party/CMakeLists.txt
@@ -10,6 +10,11 @@ if(UNIX)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC")
 endif(UNIX)
 
+if (MSVC)
+  # Disable warning about size_t conversions for all third_party libraries.
+  add_compile_options(/wd4267)
+endif()
+
 # Unit test library
 if (OpenMVG_BUILD_TESTS)
   add_subdirectory(CppUnitLite)


zwn:
This implements https://github.com/openMVG/openMVG/pull/848#issuecomment-292204795 idea. If this is merged,  #848 can be closed.

rperrot:
I don't thing taking warning under silence is a good idea. The better idea is to work on how to remove warnings in the code not by compiler trick. (a warning is a sleeping runtime error ;-) Ignoring it may be dangerous for latter developments)

Since the introduction of uintXXX_t types, the number of warnings should have increase, this means that we have to work on how to make openMVG more safe on type (sign/width).

zwn:
@rperrot Please note also https://github.com/openMVG/openMVG/issues/844#issuecomment-292508286 - the difference between warning settings for gcc and msvc. There are so many warnings issued under msvc that it is really hard to work with. I'd suggest to have the same level of warnings enabled for both compilers.

rperrot:
Yes you're right, compilers have different levels of warnings that usually do not match. 

(BTW, I think Wextra should be used with gcc/clang in order to mimic VS warnings)

pmoulon:
Thx @zwn Merged.

PR #896, opened 2017-05-01
Add CUDA detection to the top-level CMakeLists.txt

mdaiter:
We should add a standardized CUDA detection system to the top-level CMakeLists.txt file. This will enable easy enabling or disabling of possible future CUDA bindings.


pmoulon:
Thank you @mdaiter PR merged here https://github.com/openMVG/openMVG/commit/2ae3b22f540fbff4cca4d27e80602521b47870ec

PR #899, opened 2017-05-03
[build] adding KORAL to OpenMVG

mdaiter:
Adding KORAL into the main OpenMVG repository, and ensuring that it becomes built alongside the rest of the repository.

The main goals of this are as such:

1. Ensure KORAL is built in a fail-safe manner. This means ensuring that arch and PTX code is only compiled for sm_30 and up, as to avoid conflicts when compiling intrinsics.
2. Ensure that KORAL is built *only* when CUDA is enabled. As KORAL depends on CUDA, this is crucial.

The next logical step is to integrate the `image_describer_koral.hpp` class in the `src/openMVG/features` folder.


mdaiter:
@pmoulon ?

mdaiter:
@pmoulon , could I get a review? Warning: this does have other commits from different PRs involved; therefore, I'd recommend reviewing those in order to dilute down the breadth this PR will take up.

pmoulon:
I think this PR make https://github.com/openMVG/openMVG/pull/909/ useless.
Am I right?

mdaiter:
@pmoulon if you merge this entire commit into the tree, then yes. However, it might be easier to review this PR by cutting down on the total amount of files in each PR. Increased ease of reviewing was the only reason I created two PRs.

src/openMVG/features/CMakeLists.txt
```diff
@@ -20,6 +20,9 @@ list(REMOVE_ITEM features_files_sources ${REMOVEFILESUNITTEST})
 set_source_files_properties(${features_files_sources} PROPERTIES LANGUAGE CXX)
 add_library(openMVG_features ${features_files_sources} ${features_files_headers})
 target_link_libraries(openMVG_features fast)
+if (${OpenMVG_USE_CUDA})
```
pmoulon: Please do directly
`if (OpenMVG_USE_CUDA)`

src/openMVG/features/image_describer_latch.hpp
```diff
@@ -0,0 +1,160 @@
...
+  @brief Detect regions on the image and compute their attributes (description)
+  @param regions The detected regions and attributes (the caller must delete the allocated data)
+  @param mask 8-bit gray image for keypoint filtering (optional).
+     Non-zero values depict the region of interest.
+  */
+  bool Describe(const image::Image<unsigned char>& image,
+    std::unique_ptr<Regions> &regions,
+    const image::Image<unsigned char> * mask = nullptr ) override
+  {
+		std::vector<Keypoint> kpts;
```
pmoulon: indent

src/openMVG/features/image_describer_latch.hpp
```diff
@@ -0,0 +1,160 @@
...
+    std::vector<uint64_t> descriptors;
+
+    constexpr uint8_t KFAST_thresh = 60;
+    koral_.go(image.data(), image.cols(), image.rows(), KFAST_thresh);
+    kpts = koral_.kps;
+    descriptors = koral_.desc;
+
+	Allocate(regions);
```
pmoulon: indent (2 spaces large)

src/openMVG/features/image_describer_latch.hpp
```diff
@@ -0,0 +1,160 @@
...
+	  #ifdef OPENMVG_USE_OPENMP
+		#pragma omp parallel for
+	  #endif
+		for (size_t i = 0; i < static_cast<int>(kpts.size()); ++i)
+		{
+		  const Keypoint ptLatch = kpts[i];
+
+		  // Feature masking
+		  if (mask)
+      {
```
pmoulon: wrong indent

src/openMVG/features/image_describer_latch.hpp
```diff
@@ -0,0 +1,160 @@
...
+	}
+    return true;
+  };
+
+  /// Allocate Regions type depending of the Image_describer
+  void Allocate(std::unique_ptr<Regions> &regions) const override
+  {
+	switch(params_.eLatchDescriptor_)
+	{
+      case LATCH_BINARY:
```
pmoulon: wrong indent

src/openMVG/features/image_describer_latch.hpp
```diff
@@ -0,0 +1,160 @@
...
+        return regions.reset(new AKAZE_Binary_Regions);
+      break;
+	}
+  }
+
+  template<class Archive>
+  void serialize(Archive & ar)
+  {
+    ar(
```
pmoulon: You can do
```
ar(cereal::make_nvp("params", params_),
   cereal::make_nvp("bOrientation", bOrientation_));

```

src/third_party/koral/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
+PROJECT(koral)
+cmake_minimum_required (VERSION 3.1)
+
+IF (${OpenMVG_USE_CUDA})
```
pmoulon: use `if (OpenMVG_USE_CUDA)`

Please use lower case name for all the cmake functions

src/third_party/koral/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+  FIND_PACKAGE(CUDA REQUIRED)
+  FIND_PACKAGE(SSE REQUIRED)
+
+  INCLUDE_DIRECTORIES(${CMAKE_CURRENT_BINARY_DIR}/config/)
+
+  SET (HEADERS CLATCH.h CUDALERP.h FeatureAngle.h KFAST.h KORAL.h Keypoint.h)
+
+  SET (SRC CLATCH.cu CUDALERP.cu)
+
+  LIST(APPEND CUDA_NVCC_FLAGS "-D_MWAITXINTRIN_H_INCLUDED -lineinfo -Xptxas -v -Xcompiler -fopenmp -use_fast_math -O3 -gencode arch=compute_30,code=sm_30 -gencode arch=compute_60,code=sm_60 --default-stream per-thread")
```
pmoulon: Do we have to add fopenmp only if openMP is supported?

src/third_party/koral/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+
+  SET(CUDA_NVCC_FLAGS "${CUDA_NVCC_FLAGS} -std=c++11")
+
+  SET(CMAKE_CXX_STANDARD 11)
```
pmoulon: CXX_STANDARD is already set by the master OpenMVG cmake
see [here](https://github.com/openMVG/openMVG/blob/develop_koral/src/CMakeLists.txt#L135)

mdaiter:
@pmoulon all fixed.

mdaiter:
@pmoulon if there are no other issues, we should probably attempt to merge. This is crucial for moving forward, especially when adding:
1. Tests
2. Travis CI build configs (involving CUDA testing)

pmoulon:
Made some minor comments that I would like you to consider.
- C++11 compilation rules is already done in OpenMVG CMakeLists.txt
- a style issue for {}
- a performance issue bit copy vs. memcpy (can you check if we can use it?)

src/openMVG/features/image_describer_latch.hpp
```diff
@@ -74,71 +74,70 @@ class LATCH_Image_describer : public Image_describer
...
+          {
+            const image::Image<unsigned char> & maskIma = *mask;
+            if (maskIma(ptLatch.y, ptLatch.x) == 0)
+              continue;
+          }
+          const float scale = static_cast<float>(std::pow(1.2f, ptLatch.scale));
+          regionsCasted->Features()[i] = SIOPointFeature(scale * static_cast<float>(ptLatch.x), scale * static_cast<float>(ptLatch.y), 7.0f * scale, 180.0f / 3.14159263f * ptLatch.angle);
+          for (size_t j = 0; j < 8; j++) {
+            const uint64_t descriptor = descriptors[i * 64 + j];
+            for (size_t k = 0; k < 8; k++) {
```
mdaiter: I believe so!

src/openMVG/features/image_describer_latch.hpp
```diff
@@ -74,71 +74,70 @@ class LATCH_Image_describer : public Image_describer
...
+          // Feature masking
+          if (mask)
+          {
+            const image::Image<unsigned char> & maskIma = *mask;
+            if (maskIma(ptLatch.y, ptLatch.x) == 0)
+              continue;
+          }
+          const float scale = static_cast<float>(std::pow(1.2f, ptLatch.scale));
+          regionsCasted->Features()[i] = SIOPointFeature(scale * static_cast<float>(ptLatch.x), scale * static_cast<float>(ptLatch.y), 7.0f * scale, 180.0f / 3.14159263f * ptLatch.angle);
+          for (size_t j = 0; j < 8; j++) {
```
mdaiter: Will do.

src/openMVG/features/image_describer_latch.hpp
```diff
@@ -0,0 +1,153 @@
...
+     Non-zero values depict the region of interest.
+  */
+  bool Describe(const image::Image<unsigned char>& image,
+    std::unique_ptr<Regions> &regions,
+    const image::Image<unsigned char> * mask = nullptr ) override
+  {
+    std::vector<Keypoint> kpts;
+    std::vector<uint64_t> descriptors;
+
+    constexpr uint8_t KFAST_thresh = 60;
```
pmoulon: This threshold must be a detector parameter later on.

src/openMVG/features/image_describer_latch.hpp
```diff
@@ -0,0 +1,153 @@
...
+    {
+      case LATCH_BINARY:
+      {
+        // Build alias to cached data
+        AKAZE_Binary_Regions * regionsCasted = dynamic_cast<AKAZE_Binary_Regions*>(regions.get());
+        regionsCasted->Features().resize(kpts.size());
+        regionsCasted->Descriptors().resize(kpts.size());
+
+        #ifdef OPENMVG_USE_OPENMP
+        #pragma omp parallel for
```
pmoulon: I'm not sure that this parallel loop is worth it.

src/openMVG/features/image_describer_latch.hpp
```diff
@@ -0,0 +1,153 @@
...
+#include "third_party/koral/KORAL.h"
+#endif
+
+#ifdef OPENMVG_USE_SSE
+#include <xmmintrin.h>
+#endif
+
+#ifdef OpenMVG_USE_CUDA
+
+using namespace std;
```
pmoulon: We must avoid using namespace X in the header files.

src/openMVG/features/image_describer_latch.hpp
```diff
@@ -0,0 +1,153 @@
...
+#include "openMVG/features/image_describer.hpp"
+#include "openMVG/features/regions_factory.hpp"
+#include <cereal/cereal.hpp>
+
+#ifdef OpenMVG_USE_CUDA
+#include "third_party/koral/KORAL.h"
+#endif
+
+#ifdef OPENMVG_USE_SSE
+#include <xmmintrin.h>
```
pmoulon: This include should certainly must be in the client API (KORAL?)

src/openMVG/features/image_describer_latch.hpp
```diff
@@ -0,0 +1,153 @@
...
+  bool Describe(const image::Image<unsigned char>& image,
+    std::unique_ptr<Regions> &regions,
+    const image::Image<unsigned char> * mask = nullptr ) override
+  {
+    std::vector<Keypoint> kpts;
+    std::vector<uint64_t> descriptors;
+
+    constexpr uint8_t KFAST_thresh = 60;
+    koral_.go(image.data(), image.cols(), image.rows(), KFAST_thresh);
+    kpts = koral_.kps;
```
pmoulon: Seems like those copy are not necessary, since we can use directly the `koral_.kps` and `koral_.desc`.
If you want use short name you can use some `const auto&` alias.

saihv:
@mdaiter Just a quick question out of curiosity, how is memory management being done here on the GPU side? If the `Describe()` function is called in a loop for multiple images using KORAL, the call to `_koral_.go(image.data(), image.cols(), image.rows(), params_.KFAST_threshold_);` seems like it would result in multiple `cudaMalloc()` calls internally without a corresponding free function. 

Is this something that needs to be added, or is it already being handled elsewhere?

mdaiter:
@saihv you’ve got a point. @komrad36, thoughts? Seems like we do call cudaMalloc without the corresponding free functions in the go method.

komrad36:
From memory I believe the general idea was that KORAL is to be adapted for use in a pipelined realtime system, where you're receiving a succession of frames. So you'd allocate once at the start and just never deallocate. As I mention in the README, KORAL is just a prototype in that direction, suggesting how to combine CPU KFAST with GPU CLATCH and K2NN, so please treat it as such. If you adapt it for your application and call go() repeatedly, and you need a cudaFree, by all means, please add one. If you adapt it to process frames of the same size, you should be able to avoid repeated cudaMallocs and cudaFrees.

saihv:
Gotcha, thanks! I am currently looking into whether I can optimize the OpenMVG + KORAL pipeline to use preallocated memory and avoid repeated calls to malloc and free while processing multiple images (preset resolution) as well as memory leaks.

@mdaiter Looks like you'd already found a solution to this [in your Hamming distance matcher](https://github.com/openMVG/openMVG/blob/develop_koral/src/openMVG/matching_image_collection/gpu/LatchBitMatcher.cpp#L99) where you used cudaMemset on preallocated memory. KORAL has a slightly more complicated pipeline so I'm still figuring out if I can use something similar. If you happen to have any suggestions, I'd be happy to hear them.

PR #903, opened 2017-05-05
Speedup compilation times

bjornpiltz:
This PR tries to move as much code as possible away from header files. It specifically removes all references to cereal from public includes. Here are some results:
- Travis build time for all tests, samples and exes has gone from [1h44](https://travis-ci.org/bjornpiltz/openMVG/builds/228716984) to [53m](https://travis-ci.org/bjornpiltz/openMVG/builds/229157648).
- The real elapsed time went from 46 min to 20 min

More importantly the usability of openMVG has gone up. Before, a file containing the line `#include <openMVG/sfm/sfm.hpp>` took 29 s to compile. This is now down to 2.5 seconds.

Some of the main_*.cpp compiles significantly faster. Here are some examples:
image

PR #911, opened 2017-05-09
Forward declare cereal(speedup compilation)

bjornpiltz:
This PR moves all referencences to cereal to separate header files with the name `'<class name>_io.hpp'`. It is a different approach than https://github.com/openMVG/openMVG/pull/903, but seems to achieve a similar improvement in compilation time. ([1h48](https://travis-ci.org/bjornpiltz/openMVG/builds/230309313) -> [56](https://travis-ci.org/bjornpiltz/openMVG/builds/230325788))

Compilation time for all libs, test, examples & exes in minutes (user+sys):
image

The new way for developers to serialize openMVG classes is like this:

```
// First include the archive type(s) you need.
#include <cereal/archives/json.hpp>

// Then include the file suffixed '_io.hpp'.
#include "openMVG/cameras/cameras_io.hpp"

// Now you can serialize:
void load()
{
  std::shared_ptr<IntrinsicBase> intrinsic(nullptr);
  std::ifstream stream(filename, std::ios::binary | std::ios::in);
  cereal::JSONInputArchive archive(stream);
  archive(cereal::make_nvp("intrinsics", intrinsic));
}
```
Including 'cameras.hpp' instead of 'cameras_io.hpp' will generate the error `cereal could not find any input serialization functions for the provided type and archive combination.`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,13 +65,13 @@ before_script:
       -DCMAKE_CXX_COMPILER=$COMPILER \
       -DCMAKE_BUILD_TYPE=release \
       -DOpenMVG_BUILD_TESTS=ON \
-      -DOpenMVG_BUILD_EXAMPLES=OFF \
-      -DOpenMVG_BUILD_SOFTWARES=OFF \
+      -DOpenMVG_BUILD_EXAMPLES=ON \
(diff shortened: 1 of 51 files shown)


pmoulon:
Thank @bjornpiltz.
Changed merged!

PR #912, opened 2017-05-09
[build] add CUDA to optional dependency

mdaiter:
We should add an "Optional dependency" section detailing that users can activate or deactivate CUDA (and in turn KORAL).


mdaiter:
@pmoulon easy review ;)

Issue #797, opened 2017-02-21
Use of dynamic_cast with Regions


hugbed:
Hi,

I was looking at the [robust estimation tutorial homography](https://github.com/openMVG/openMVG/blob/master/src/openMVG_Samples/multiview_robust_estimation_tutorial/robust_estimation_tutorial_homography.cpp) and couldn't help but notice the use of a dynamic_cast to a derived class (`SIFT_Region`) to retrieve function members such as `Descriptors`. 

Wouldn't there be an alternative to `dynamic_cast` since in a real-time applications, the non-deterministic dynamic casts could be problematic? From what I have learned, dynamic casts are used sometimes because the interface cannot correctly represent the implementation (maybe I'm wrong). Maybe the Image_describer could be templated with its corresponding `Regions` type, e.g., `SIFT_Image_describer ` could derive from `Image_describer<SIFT_Regions>` and use the templated type as region.

I don't know if it's just me but it feels weird to use dynamic_cast as a standard way to use an interface.

For example in Image_describer (c++ pseudo-code):

```cpp
template <class RegionType = features::Region>
class Image_describer
{
  // ...
}
```

And SIFT_Image_describer:

```cpp
template <class RegionType = SIFT_Regions>
class SIFT_Image_describer<RegionType>: Image_describer<RegionType>
{
  // ...
}
```

There could be different template specialization for different region types to replace the switch cases in the implementation. However, this seems like much work and would change the current interface...

Something like that could render the Region type available at compilation so the `dynamic_cast` would not be necessary would be great:

```cpp
std::unique_ptr<Image_describer<SIFT_Regions>> image_describer
    (new SIFT_Image_describer(SIFT_Image_describer::Params(-1)));
std::map<IndexT, std::unique_ptr<SIFT_Regions> > regions_perImage;
image_describer->Describe(imageL, regions_perImage[0]);
image_describer->Describe(imageR, regions_perImage[1]);

SIFT_Regions *regions = regions_perImage.at(0).get();
```

I'm not sure about how this could be implemented but I would like your opinion on the matter!

Thank you.

pmoulon:
Hi @hugbed 
Thank your for your feedback. Here my answer.

1. I have choose the design to be as abstract as possible because for example a single Image_Describer can return various regions type and the major point is that I would like to write generic code.

If you take as example the AKAZE_Image_describer ([here](https://github.com/openMVG/openMVG/blob/master/src/openMVG/features/image_describer_akaze.hpp#L109))
[...]

PR #913, opened 2017-05-09
Develop image describer covariant

hugbed:
## Why these changes

I was looking at the [robust estimation tutorial homography](https://github.com/openMVG/openMVG/blob/master/src/openMVG_Samples/multiview_robust_estimation_tutorial/robust_estimation_tutorial_homography.cpp) and couldn't help but notice the use of a dynamic_cast to a derived class (`SIFT_Region`) to retrieve function members such as `Descriptors`.

I wondered if there could be a small design change to prevent the use of dynamic_casts that have a non-deterministic cost.

The original issue is here #797 that details my thoughts at the time.

## What I Changed

The Image_Describer interface was modified so that no `dynamic_cast`s are needed inside a the `Describe` function of a derived class. This was achieved by returning a `unique_ptr<Regions>` instead of passing a reference to a unique_ptr<Regions> to the Describe function. No dynamic_cast should be needed when using the `Regions` returned by a specific class since specific classes now return specific `Regions`.

The virtual `Describe` methods are now private and return the generic type.
The public method provides the specific type through a non-virtual interface (for smart pointer ''covariance''). Returning the `Regions` allows us to use covariance to return a specific type for a specific derived class and a generic type for the base interface. It can prevent the use of `dynamic_cast` in `Describe` function implementations and when using the derived class. Such covariance is not possible with function parameters. 

Since every derived class should be able to implement the generic `Describe` method, this is the one that is virtual. The specific `Describe` method (returning a specific `Regions_type`) is specific to the derived class and is then non-virtual.

This change uses the move semantics of C++11. Local variables (here it's a `unique_ptr<Regions>`) can be returned with minimal cost.

*BEFORE*:

```cpp
class Image_Describer {
public:
  virtual bool Describe(
        const image::Image<unsigned char> & image,
        std::unique_ptr<Regions> &regions,
        const image::Image<unsigned char> * mask = nullptr
  ) = 0;
};
```

*AFTER*:

```cpp
class Image_Describer 
{
public:
  std::unique_ptr<Regions> Describe
  (
    const image::Image<unsigned char> & image,
    const image::Image<unsigned char> * mask = nullptr
  )
  {
    return DescribeImpl(image, mask);
  }

  // ...

private:
  virtual std::unique_ptr<Regions> DescribeImpl(
    const image::Image<unsigned char> & image,
    const image::Image<unsigned char> * mask = nullptr
  ) = 0;

  // ...
}
```

Using `Describe` from the base class will return the generic type through the base class generic `Describe` method that calls the specific type. Using `Describe` from a derived class will return the specific type through the specific `Describe`, e.g :

*BEFORE*:

```cpp
class SIFT_Image_Describer {
  bool Describe(
        const image::Image<unsigned char> & image,
        std::unique_ptr<Regions> &regions,
        const image::Image<unsigned char> * mask = nullptr
  ) override
  {
    Regions* regions = Allocate();
  }
};
```

*AFTER*:

```cpp
class SIFT_Image_Describer
{

  using Regions_type = SIFT_Regions;

public:
  std::unique_ptr<Regions_type> Describe
  (
    const image::Image<unsigned char> & image,
    const image::Image<unsigned char> * mask = nullptr
  )
  {
    // ...

    auto regions = Allocate();
    // fill regions
    return regions
  }

  // ...

private:
  virtual std::unique_ptr<Regions> DescribeImpl(
    const image::Image<unsigned char> & image,
    const image::Image<unsigned char> * mask = nullptr
  ) override
  {
    // this works since you can create a unique_ptr<Regions> from a unique_ptr<SIFT_Regions>
    return Describe(image, mask); 
  }

  // ...
}
```

`Allocate` was also modified to have smart pointer covariance with a non-virtual interface. Since allocate allocates the memory, it transfers the memory ownership to the user and a smart pointer should be used. Having a non-virtual interface allows covariance properties with smart pointers.

```cpp
class Image_Describer 
{
public:
  virtual std::unique_ptr<Regions> Allocate()
  {
    return AllocateImpl();
  }

  // ...

private:
  virtual std::unique_ptr<Regions> AllocateImpl() const = 0;

  // ...
```

Calling `Describe` with a unique pointer to Regions by reference is still possible as before:

```cpp
bool Describe
(
  const image::Image<unsigned char> & image,
  std::unique_ptr<Regions> &regions,
  const image::Image<unsigned char> * mask = nullptr
)
{
  regions = DescribeImpl(image, mask);
  return regions != nullptr;
}

```

With this change, a class derived from `Image_Describer` should return only one type, which reduces the number of switch cases and enforces polymorphism instead. This is why `AKAZE_Image_Describer` was separated in 3 different classes (SURF, LIOP, MLDB).

To create a specific type of `AKAZE_Image_Describer`, I added a static create function to `AKAZE_Image_Describer` that takes the same parameters as the constructor and returns a new specific `AKAZE_Image_Describer` from the chosen `Regions_type`. Maybe having this function in a factory would be a better idea however since only this class has this create function for now which may be confusing to have multiple ways of creating Image_Describers. It's also weird that the base class knows about its derived classes.

```cpp
std::unique_ptr<AKAZE_Image_describer> AKAZE_Image_describer::create
(
  const AKAZE_Image_describer::Params& params,
  bool orientation
)
{
  switch (params.eAkazeDescriptor_) {
  case AKAZE_MSURF:
    return std::unique_ptr<AKAZE_Image_describer>(new AKAZE_Image_describer_SURF(params, orientation));
  case AKAZE_LIOP:
    return std::unique_ptr<AKAZE_Image_describer>(new AKAZE_Image_describer_LIOP(params, orientation));
  case AKAZE_MLDB:
    return std::unique_ptr<AKAZE_Image_describer>(new AKAZE_Image_describer_MLDB(params, orientation));
  default:
    return nullptr;
  }
}
```

The tests still pass, I only had to modify the creation of `AKAZE_Image_Describer` to create specific describers.

Developpers who implement the `Image_Describer` interface may have to modify their code so that it matches the new interface. However, this should remove some dynamic_casts in the derived class so it can be a good thing.

## Pros and cons

Pros:
* No more dynamic casts when the `Regions` type is known : efficiency and better coding practices.
  * No dynamic cast when implementing `Describe` to recover lost `Regions` type
  * No dynamic cast when using a specific `Image_describer`
* `Allocate` now returns a `unique_ptr` since it gives up its ownership of the memory allocated.

Cons:
* More boilerplate code, 2 functions instead of 1 : 1 virtual and 1 non-virtual for smart pointers
* Non-virtual methods hide the base class methods, it could be confusing that a derived class behaviour is reached from a non-virtual method


bjornpiltz:
I like the idea to simplify the user experience. (And I love the way you documented the PR!) My only question is if it can't be done easier like this:

``` C++
class Image_Describer 
{
public:
  virtual std::unique_ptr<Regions> Describe(...) = 0;
};

class SIFT_Image_Describer
{
public:
  std::unique_ptr<SIFT_Regions> DescribeSIFT(...)
  {
    // ...
  }

  virtual std::unique_ptr<Regions> Describe(...) override
  {
    return DescribeSIFT(...);	
  }
};
```

This way a user who is actually interested in getting a SIFT region has to call `SIFT_Image_Describer::DescribeSIFT()` and some one who doesn't care about which kind of region calls `Image_Describer::Describe()`.

I don't see the benefit of having a function named `Describe()` sometimes return this and sometime that. Also as you mentioned in your cons, [Never redefine an inherited non-virtual function](http://intercontineo.com/article/4250256884/)

hugbed:
@bjornpiltz 
I like that. Indeed, this would solve most of the cons :
* The return type of the method would always be clear has you stated
* This would prevent having a non-pure virtual function in `Image_describer` interface only to call the derived class `Describe`.
* This would remove the additional `DescribeImpl` method in `Image_describer`.

For the `Allocate` method, an alternative could be to define an `owner<T>` to use raw pointer covariance while documenting the expected behaviour. See [CPPCoreGuidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#SS-views):

```cpp
template <class T>
using owner = T;

class Image_describer {
public:
  virtual owner<Regions*> Allocate() = 0;
}

class SIFT_Image_describer {
public:
  owner<Regions*> Allocate() {
    return new SIFT_Regions(...);
  }
```

This would state in the code that the caller is meant to take ownership of the pointer and must delete it afterwards. Like using raw pointers but with documentation in the code.

bjornpiltz:
@hugbed Or, just let `Allocate()` return a `std::unique_ptr<Regions>`. I don't see the benefits of a covariant return type here. In practice, the only time you will need the function Allocate() is when you don't know which type you are dealing with. If you know you need an empty  `SIFT_Regions`, you can just call `new SIFT_Regions`. There is no reason to call `Allocate()` in `Describe()`.

hugbed:
@bjornpiltz Yes, indeed. `Allocate()` is not necessary when you know the type you want to allocate.
Now that you mention it, a plain new would be better inside `Describer()` implementations and there would be no use of this confusing non-virtual confusing `Allocate()`.

hugbed:
Maybe a bit out of subject but I was wondering what were the coding standards for indentation and such while making the changes. I tried following the standards I deduced from the files I read but there seems to be inconsistencies. Maybe having a .clang-format would help! If so, I'm trying to make one that follow most of the conventions that seem to be followed. I could propose it in another pull request if you think it could be useful.

pmoulon:
We use 2space indentation and then we focus on readability.
I will work on a code style page and .clang-format and any help is welcomed.

You can interact here about it:
https://gist.github.com/pmoulon/71d2c8e82a2b1735707d72b0223e6093

hugbed:
@bjornpiltz
I have made the changes you suggested.
This was indeed a cleaner alternative: less code and probably less confusing.

Recap:

```cpp
class Image_Describer 
{
public:
  virtual std::unique_ptr<Regions> Describe(...) = 0;
  virtual std::unique_ptr<Regions> Allocate() = 0;
};

class SIFT_Image_Describer : public Image_Describer 
{
public:
  using Regions_type = SIFT_Regions;

  std::unique_ptr<Regions_type> DescribeSIFT(...)
  {
    // ...

    auto regions = std::unique_ptr<Regions_type>(new Regions_type);

    // fill regions ...

    return regions;
  }

  virtual std::unique_ptr<Regions> Describe(...) override
  {
    return DescribeSIFT(...);	
  }

  std::unique_ptr<Regions> Allocate() const override
  {
    return std::unique_ptr<Regions_type>(new Regions_type);
  }
};
```

pmoulon:
Thank you @hugbed.
Merged by https://github.com/openMVG/openMVG/commit/becdfd3b48917b7b24ad5623d0104ed6a87fdd05.

Thank you for the detailed description related to the PR. Really appreciated!

At the end we have:
- less dynamic_cast,
- Image_describer::Allocate implementation is simpler.

hugbed:
Great! This actually is my first contribution to open source. I'm glad you appreciate my work!

PR #916, opened 2017-05-12
Created library geometry. 

bjornpiltz:
Moving heavy duty Eigen calculations away from header files should speedup compilation for files including geometry.


pmoulon:
Thank you @bjornpiltz. Merged!

Here the timing (make clean; time make -j8):
- Before:
  -  real 5m28s user 35m57s
- **After (geometry_library):**
  -  real 4m55s user 32m55s

PR #926, opened 2017-05-20
Added camera sensor width for Motorola G4 Plus

manmohanbishnoi:
Added camera sensor width for Motorola G4 Plus

Reference: http://www.devicespecifications.com/en/model/54ab3c09

Tested this with photos taken from my Moto G4 Plus phone. Works good. The pictures had EXIF data as

    Make: Motorola
    Model: Moto G (4)


pmoulon:
Merged by https://github.com/openMVG/openMVG/commit/e17cd641f97648fedaed723a1bd5536cccf26bc5.
Thanks.

PR #934, opened 2017-05-26
Connect the CamModel with CamBrand

chaoer:
Some photo exif do not cantain cambrand name in cam model.
To ensure we can handle with those images, we must connect  Camera Brand with Model


rperrot:
I don't know what to think about that. 

Is it an issue that openMVG should handle ? What I mean is that : 
1) Is the error is with the way openMVG considers the naming ? 
2) Is the error caused by an incorrect way to import the image (from the camera to the computer), or with image processing software that touch/remove/modify exif infos ? 

If it's 1) I'm ok with suggestion, if it's 2) I disagree (openMVG can do a lot but not fix what other software breaks). 

Note : the two links are broken.

chaoer:
@rperrot 
Thanks for your reply.
I also don't know this error occurs because of which reason ( just as you list). 
I just found something wrong when doing experiments and made this commits. So maybe it is better that I close this pull request.

PS:
I don't know why when directly click the two links the cannot redirect to the right place. I try to copy the url and paste to chrome, it is fine to link to the right page.

rperrot:
Maybe you don't have to close the PR. 

Since you have made the code, maybe we it can be a fallback mode if standard way of finding infos fails. (@pmoulon, an idea ?)

chaoer:
@rperrot 
Unh. I will keep this change in my dev branch. If anyone else also encounter the situation I meet, I would be happy to share this with them and reopen this pull request. : )

pmoulon:
Thanks @chaoer. 

This case is tricky since all the manufacturer does not make the same choice for storing the exif data.

So, as @rperrot state, we have to choose if we want to handle this "error" in our side or put a message to the user explaining that the data are not in the desired format.

chaoer:
@pmoulon  
Yeah, I completely agree with you.

"Put a message to the user " may the the best choice. 
Just as @rperrot state, it's really not the responsibility to handle this "error".

PR #943, opened 2017-06-06
[Build] Add skylake support

rjanvier:


pmoulon:
Perhaps we can also test to sync with https://github.com/saihv/openMVG/commit/c8916fed4bae31cce300c661d2018d2669ed2b8a

rjanvier:
Well I guess the one from @saihv is better. Maybe, it could directly be merged in the develop branch.

pmoulon:
my concern is about the stability of this new changes.
I mean if we merge the new script we won't be able to check every combination.
But perhaps we can trust it if it works on all our computers...

rjanvier:
For now, the update from @saivh, works like a charm on my skylake based computer on windows. It trully fix the detection of AVX2 in this case. My proposal was solely detecting AVX.

saihv:
FYI: There was a specific addition I had to make in my version of the cmake file to get it to work with AVX2. I had to add the AVX2 flags to both CXX and C flags (see line 518) without which I was having errors with smmintrin.h and other AVX2 command related include files, which are called from KORAL. Would be great to test it with different architectures and the KORAL branch which uses a lot of AVX2.

pmoulon:
@rjanvier I think that this problem is already solved in OpenMVG v1.2 with this commit:
https://github.com/openMVG/openMVG/commit/4c711c6bbb90fa3f780806ae44002a836e346081

Can you check that OpenMVG v1.2 is working for your architecture and close this PR if the answer is yes?

PR #953, opened 2017-06-21
[docker]: add CUDA GPU support

mdaiter:


mdaiter:
Ah. I realized that the Nvidia CUDA project actually is a far better abstraction for this. We should use that instead of this.

mdaiter:
Closing this now for history's sake.
Here's a link to Nvidia CUDA docker images: https://github.com/NVIDIA/nvidia-docker

And from there, we should use this command: ```nvidia-docker pull nvidia/cuda``` to pull in a cuda branch from Nvidia.

From there, we can easily add in CUDA support to our apps. This is a far cleaner solution, as only one layer can control GPU access at a time in a hypervisor-like system such as Docker.

More info here: https://devblogs.nvidia.com/parallelforall/nvidia-docker-gpu-server-application-deployment-made-easy/

PR #961, opened 2017-07-03
Local origin

hcjghr:
Hi @pmoulon,

this small PR is regarding the origin movement when using gps coordinates. It adds:
- a binary (main_ChangeLocalOrigin) to move the reconstruction to new local origin (either provided with -l switch or equal to the first camera position (-f))
- a normalization and denormalization step in the sequentialSfM which translates the priors to the computed center of all the priors in order to better condition the problem

Hope it is useful!

Best,
Klemen


hcjghr:
I removed the normalization I added as I found that it is already done. Sorry for the mess. Now the only addition of this PR is the binary for translating the reconstruction.

pmoulon:
Thanks, here some comments:
Can you remove the two commit related to the file `src/openMVG/sfm/pipelines/sequential/sequential_SfM.cpp` and `src/software/SfM/main_IncrementalSfM.cpp`.
This way you will have only commit related to the new binary.

Can I suggest to use the existing function `ApplySimilarity` to move the sfm_data (https://github.com/openMVG/openMVG/blob/master/src/openMVG/sfm/sfm_data_transform.hpp)?

- If you set `transform_priors = true` it can perform the same task as the one you do. It will simplify the code to the creation of the motion (where only the translation will be set) and calling the `ApplySimilarity` function ;-).

src/software/SfM/main_ChangeLocalOrigin.cpp
```diff
@@ -0,0 +1,189 @@
...
+    sOutDir,
+    sLocalFrameOrigin;
+    
+  Vec3 local_Frame_Origin;
+  bool b_first_frame_origin;
+
+  cmd.add(make_option('i', sSfM_Data_Filename_In, "input_file"));
+  cmd.add(make_option('o', sOutDir, "output_dir"));
+  cmd.add(make_option('l', sLocalFrameOrigin, "local_frame_origin"));
+  cmd.add( make_switch('f', "first_frame_origin") );
```
pmoulon: `(make` for consistency with the upper line remove the useless spaces

src/software/SfM/main_ChangeLocalOrigin.cpp
```diff
@@ -0,0 +1,189 @@
...
+    }
+  }
+    
+  // Move view priors
+  {
+    for (auto & view : sfm_data.GetViews())
+    {
+      if (sfm::ViewPriors *prior = dynamic_cast<sfm::ViewPriors*>(view.second.get()))
+      {
+	    if (prior->b_use_pose_center_)
```
pmoulon: tab inconsistency. But as I said in the general comments, I think it can be better to use the `ApplySimiliarity` function ;-) to apply a given transform (here a translation) to the whole sfm_data object.

pmoulon:
Regarding your idea, it could be nice to save the applied transform as a JSON object.
So using the same binary you can move the origin back to the original domain by using the inverted transform.

I.e:
`openMVG_main_ChangeLocalOrigin -i .. -o .. -f`
-> it can export a file sfm_data_transform.json
then you can call 
`openMVG_main_ChangeLocalOrigin -i ... -o ... -t sfm_data_transform.json --inverted`
it can read the affine transform and then apply the reverse.

We can add this functionality in another PR. What do you think about it.

erikreed:
I tried out this PR (worked great) and noticed it solves the discretization issues when viewed in Meshlab. The same issue was described in https://github.com/openMVG/openMVG/issues/949#issuecomment-309927279. The following two figures are from master branch with the fix from https://github.com/openMVG/openMVG/issues/949#issuecomment-310429812 applied.

Before:
image


After offsetting to first frame:
image

I wonder if this is due to the ply truncation applied at https://github.com/openMVG/openMVG/blob/6ce820b2844a98941a5d2543050f4e278a323800/src/software/SfM/SfMPlyHelper.hpp#L46, which is then mitigated by the few digits of extra space from storing offsets instead of ECEF coordinates, as shown by the delta from using this PR.

pmoulon:
1- I don't think its a truncation error, I'm exporting double with full precision.
BTW, I made a tool to make a fast check of the GPS trajectory of a image sequence. It can allow to check very fast if you have truncation or inaccurate trajectory (noise).

2. Regarding the discretization issue, it happens sometimes in the BA stage. I did not explain it clearly yet, I discuss about it with Ceres team and they think the problem can be due to the viewer tool.
- we can try to center the scene in (0,0,0) before exporting in order to see if we still observe the discretization.
- if the problem is within the BA we can try to save the input sfm_data scene just before the BA and share it here in order to have many people look to what is happening.
Note: Everytime I got this trouble, settings no loss for the first BA solved the issue.

jpwhitney:
Awesome work!  This seems extremely useful for UAV footage.

It would be great to be able to build this binary into a UAV pipeline script that can make use of the georeferencing geodesy binary.

Is something like this possible where the construction would be georeferenced, recentered, MVS performed and texture applied, then translated back into georeferenced coordinates?  Essentially I'm asking if the order I outlined is correct and if openMVG has a tool to apply the inverse transform to the model output by openMVS.

rjanvier:
A quick hint, @jpwhitney you can apply any 4x4 transformation matrix in [CloudCompare](www.danielgm.net/cc/). you can do all your MVS computation in a local arbitrary frame and then apply the 4x4 Matrix in CC to transform your final product. Moreover CC handle big coordinates either by compiling it with 64 bit support or by using the [global shift mechanism](http://www.cloudcompare.org/doc/wiki/index.php?title=Edit_global_shift_and_scale).

jpwhitney:
Thanks for the input @rjanvier I'll look into the CC command line syntax so I can avoid using a GUI for this task.

hcjghr:
Hi,

I noticed this "truncation effect" when showing the results using meshlab. As you already saw its not a loss of precision because you can still correctly show the result using Cloud Compare or maybe some other tool capable of handling big coordinates. 

This binary supports the change of origin through -l "x;y;z" where you define which coordinate should be the new origin or using -f switch which automatically moves the first  camera to 0,0,0. For me the second options comes handy as I dont have to define a local origin by hand every time I build a new model. The process also creates a txt file where the location of the new (0,0,0) is saved so you can easily invert the move at the end and keep the georeference information. We could easily save this as an transformation matrix as well to make the process even easier later if you find it useful. 

The final result as @rjanvier metioned can easily be moved back to the original coordinate system using cc.

cheers,
Klemen

mchildsCO76:
I am attempting to use the openMVG_main_ChangeLocalOrigin application to solve the same precision issue with GPS ECEF coordinates. Could you clarify where in the pipeline you call the ChangeLocalOrigin function?

For example, if I'm using a pipeline like:

(1) openMVG_main_SfMInit_ImageListing -P -W w0 w1 w2 ...
(2) openMVG_main_ComputeFeatures ...
(3) openMVG_main_ListMatchingPairs -G -n limit_num_of_pairs -o match_dir/pair_list.txt ...
(4) openMVG_main_ComputeMatches -l match_dir/pair_list.txt ...
(5) openMVG_main_IncrementalSfM -P ...
(6) openMVG_main_geodesy_registration_to_gps_position
(7) openMVG_main_openMVG2openMVS -i recon_dir/sfm_data.bin -o recon_dir/scene.mvs -d ./undisto
(8) DensifyPointCloud -i recon_dir/scene.mvs -o recondir/scene_dense.ply

Would I just do the change local origin before step (5) [the incremental or global SfM call], then restore afterwards, or should I change the origin immediately after (1) and keep it that way through the entire process, then manually apply the offset back to the generated PLY when I read it in as a point cloud?

I can probably figure this out with some playing, but given how long each run of the pipeline takes on my data set if you have a quick answer that would be helpful.

hcjghr:
The problem doesnt really occur before (5) as nothing is reconstructed up to that point (all the information you have is about 2D features and their matches between images). The Incremental BA (5) at the beginning actually applies a transformation on the GPS positions so that  the center of the reconstruction will be in (0,0,0) and at the end applies the reverse of that to get properly georeferenced solution.  Given that (6)geodesy_registration_to_gps_position still assumes the "global" GPS position I would not move the local origin before this step. My advice would be to do it just  before exporting to openMVS (openMVG is able to cope with big coordinates so there is no reason to complicate your life:)). 

And then once the whole process (openMVS + if you have anything else) is completed you will have a model that its (0,0,0) is actually located in the position exported in a txt file by change_local_origin and you can apply the reverse transformation (just translation).

I hope my reasoning makes sense.

mchildsCO76:
Thank you your reasoning worked perfectly! I am now getting much better results for point cloud generated from drone imagery!

pmoulon:
We need to work on integrate this PR and update the doc there:
https://github.com/openMVG/openMVG/blob/master/docs/sphinx/rst/software/Geodesy/geodesy.rst

- we can use some c++11 range loop in order to make the code simpler to read
- we have to fix some tiny minor code style issues (alignement)

mchildsCO76:
One more question on this... I get a much better spread of XY points after doing the local origin change and everything seems quite precise. The Z coordinates however are spread a bit more than I'd expect. The resulting point cloud (after converting from the local system back to ECEF and then to lat/lon with elevations in meters) is giving me a spread of about 0.3-0.4m for surfaces that should be flat (like a road). This happens with multiple data sets.

I suspected maybe the lack of precision in the Z coordinate in the EXIF location data embedded in the JPG files from the drone GPS, but point clouds for the same data from other software (AGIsoft for example) do not have this thickness issue.

Have you seen something similar in your work with the local coordinate offset? If not I can move this to a new thread.

mchildsCO76:
Here's an example showing a generated point cloud with a cross section 10 cm wide across a road. You can see that the points for the road surface are about 0.5m thick.

gm_screenshot

pmoulon:
@mchildsCO76 Do you have any dataset (or a subsection) to share in order to help us to check what's going on?
If you don't want to share the images you can only share the match directory without the *.desc files. It should be quite small once zipped.

mchildsCO76:
I've place a .zip with the images as well as a cmd_list.txt listing each command line that was run during the process at ftp://BMGAug17:project@ftp.bluemarblegeo.com/Mike/mike_niwot_house_drone_data.zip . There is also an image_to_cloud_log.txt lists the output of all of those commands as well as extra logging of the image info at the start. The local camera database is also there in case the camera in those images (from a DJI Mavic drone) was one automatically added at runtime from the EXIF info and not one in the list from the repository.

rttgnck:
@mchildsCO76 This is also something I am running into. Have great data, but not as flat as it could be. I'm seeing the same dispersion in the point cloud when viewed in a cross section. I'd really like to talk to you some more as we appear to be working on the exact same end product.

mchildsCO76:
I have noticed that if I run the same process but with Incremental SfM rather than Global SfM I get a smaller cross section (0.2m rather than 0.5m or more). I'm not entirely sure of the advantages/disadvantages to using Incremental vs. Global SfM, but will do some more reading.

rttgnck:
@mchildsCO76 in issue #1037 I asked Pierre about just that and now he would like to do a FAQ about it soon. 
So if you use Incremental you get a smaller variance than Global. Which means it should be able to be manipulated. I suspected it was in the MVS DensifyPointCloud, not MVG. Looks like something generated in the MVG process that is used in DensifyPointCloud that is resulting in the difference. At work now, so I can't look into this any further until later. 
Are you processing using Incremental vs Global the same, with GPS exif, or are you letting the Incremental stage process on its "everything is no larger than 1" scale? I assume the same process just different pipeline, so the variance while still there is just different between the two so there is a tolerance of some kind that differs between the two, or the tolerance is the same in each and the difference in processing between the pipelines is what causes the variance.
I will have to experiment with this myself later. I have done most of my processing in the Global pipeline.

mchildsCO76:
I am doing the same processing with both Global and Incremental, using GPS EXIF if present. It was my initial expectation that Global would generate better results, but it seems Incremental is the way to go unless there is some other option that I'm missing.

rttgnck:
That's likely the case. I am looking at the same situation right now.
CloudCompare has a noise filter and it helped mine, but only helped, didn't
fix. On a 8.7m point cloud I ran it 3 sequential times and got the surface
down to about 2m points. I guess if it lowers the number of points to a
still usuable amount, then it might be more beneficial for a mesh
reconstruction if there are less points. On the other hand I want a dense
and perfect point cloud on the order of 20m points, but that's just me.

pmoulon:
Do you have any idea about your GPS quality?
- some registration statistics for the GPS fitting are printed out during the SfM process. It could be interesting to see those value or see how we can enhance the SfM report to have a better feedback about it.

pmoulon:
I read the SFM log you share me
```
Registration report:
 #corresponding SFM - GPS data: 109

Found transform:
 scale: 0.990221
 rotation:
    0.999999  -0.00141793 -0.000753359
   0.0014195     0.999997   0.00207995
 0.000750407  -0.00208101     0.999998
 translation: -16032.352110827 -52878.945410045  31040.406142883

3D Similarity fitting error (in target coordinate system units):
	 min: 0.073266529
	 mean: 0.596290469
	 median: 0.615372121
	 max: 1.137230992
```

So it looks like the median error between your camera poses and GPS data is about 0.615372121 meters.

rttgnck:
@pmoulon I just looked at a report from my data and see a ~0.45m mean error as well.

If that is the case then the only option for better results is a more accurate GPS system? That's not really possible with the Mavic Pro.

My limited understanding is that the GPS signal is accurate to about 2-3 meters in Z. That it takes approx 2s per photo in JPG to save, and 10s per DNG (raw) to save. The issue there is DJI had some time back switched from the barometer to the GPS altimeter reading in the JPG exif. While the barometer is more accurate for distance above ground and people report it tends to stay more consistent than GPS on a 20 minute job, however the time it takes to save the photos means the jobs could take up to 5x as much time to complete. It would be really nice if DJI added the option to store either or, or both, in the JPG exif. I mean come on, they used to use the barometer data and switched.

pmoulon:
[Here](https://pix4d.com/rtk-ppk-drones-gcp-comparison/) an interesting read that shows some accuracy results for field survey for the GPS-RTK (Fixed/Float) case. You will see that they have the same problem (accuracy is different along x,y and z axis).

pmoulon:
@mchildsCO76 
Can you explain me what is the purpose of the `-x` option you choose to try on `openMVG_main_SfMInit_ImageListing`. This option does not exists. Did you tried to intend to use `-c 1` instead?

mchildsCO76:
That's the extended camera name lookup option discussed at https://github.com/openMVG/openMVG/issues/1002 for matching a partial camera name. I have that option added in a fork, but no PR yet or anything. It turns out I don't actually need it with how I manage my camera information now (I build my own camera file on-the-fly with the exact entry) so it might end never being added. It was pretty insignificant.

pogilon:
I don't seem to find the "openMVG_main_ChangeLocalOrigin", in which branch is it? Did you change its name?

pogilon:
Any news about when to include this RP?

src/software/SfM/main_ChangeLocalOrigin.cpp
```diff
@@ -0,0 +1,189 @@
...
+	    {
+          prior->pose_center_ = prior->pose_center_- local_Frame_Origin;
```
pmoulon: ` ` - ` `

pmoulon:
Thanks @hcjghr 
Merged by https://github.com/openMVG/openMVG/commit/5429d9ba1a77b08568cf5dc5f11064119e094577 and https://github.com/openMVG/openMVG/commit/374bc69745295ad241c46038e0470142799de78a

PR #962, opened 2017-07-04
[bug fix] main_ComputeFeatures multi-threading issue.

linfan68:
Fixed the issue that when running multi-threaded different threads will keep overriding the same image data.


pmoulon:
Thanks.
This bug is fixed in the develop branch and so will be included in the next coming version.

pmoulon:
The issue is now solved since OpenMVG v1.2 have been pushed to master.

PR #1060, opened 2017-09-15
Control Points Registration GUI fixes and improvements

trylle:
Pull request, as requested. I made some minor additional changes:

* GraphicsMainWindow has been renamed GraphicsView and now inherits from QGraphicsView instead of QMainWindow. This negates having to create a separate class to handle the mouse wheel events properly. I am assuming the original supertype was not selected for a specific reason, as it is unusual to have nested QMainWindows. Having the QGraphicsScene in this class now makes it more apparent that model-view isolation is being broken, but that's something that can be resolved later on if needed and no worse than it already was, I suppose.
* Fixed an ambiguous shortcut between save and normal size.


rperrot:
You're right, graphics view should only be used for rendering a view, it may not derive from QMainWindow. 

Since you made some changes, I suggest you to help us enhance more this GUI : 
- File naming is inconsistent : file should reflect the class it holds (for ex: mainLayout.hpp should be renamed to MainWindow.hpp);
- Members formatting is inconsistent (each member should start with a lower case - for ex AddImage should be addImage in GraphicsView.hpp) ;
- #pragma once should be removed and replaced to #ifndef / #define clauses ;

trylle:
I believe I've now made the needed changes in the spirit of your remarks.

pmoulon:
Thank you very much @trylle 
PR squashed and merged thanks to https://github.com/openMVG/openMVG/commit/09923fbbd03a1043add7be616af31700a17dd032
Looking forward other contributions soon ;-)

PR #1069, opened 2017-09-28
Fix broken link in readme

bjohnson-va:


pmoulon:
Hi @bjohnson-va,
Thank you for the fix!
It will be pushed and merge in the next release.
Looking forward to see you back with other contribution or some screenshot of reconstruction on G+ #openMVG  ;-)

PR #1091, opened 2017-10-09
Add Incremental Matching

luca-morreale:
Hello,
I add the possibility to compute the matches incrementally.
It starts from a set of matches, otherwise it just computes all of them, and from them it computes the missing ones.
A match is computed only if at least one of the two views involved is not present in any of the pre-computed match.


pmoulon:
Seems like it does not compiles on Unix:
https://travis-ci.org/openMVG/openMVG/jobs/285531810#L1921

luca-morreale:
You are right, I actually forgot to commit a change.
Now it should compile properly.

pmoulon:
No worries, the continuous integration is here to check that (safe compilation and unit test launching).

Issue #1081, opened 2017-10-05
SfMWebGLViewer: compile error gcc 7.2 (Fedora 26)


sgielkens:
Bonjour Pierre,

First of all, thanks for the continuously improving project!

I checked the open issues but I did not see the following issue. A compile error occurs in src/software/SfMWebGLViewer/openMVG_main_openMVG2WebGL.cpp:

`/builddir/build/BUILD/openMVG-master/src/software/SfMWebGLViewer/openMVG_main_openMVG2WebGL.cpp:381:98: error: 'OPENMVG_SFM_WEBGL_SRC_PATH' was not declared in this scope
   const std::string inputFilename1 = stlplus::create_filespec( stlplus::folder_append_separator( OPENMVG_SFM_WEBGL_SRC_PATH ) + inputFolder , inputName );
                                                                                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~
/builddir/build/BUILD/openMVG-master/src/software/SfMWebGLViewer/openMVG_main_openMVG2WebGL.cpp:381:98: note: suggested alternative: 'OPENMVG_SFM_SFM_VIEW_HPP'
   const std::string inputFilename1 = stlplus::create_filespec( stlplus::folder_append_separator( OPENMVG_SFM_WEBGL_SRC_PATH ) + inputFolder , inputName );
                                                                                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~
                                                                                                  OPENMVG_SFM_SFM_VIEW_HPP
/builddir/build/BUILD/openMVG-master/src/software/SfMWebGLViewer/openMVG_main_openMVG2WebGL.cpp:382:98: error: 'OPENMVG_SFM_WEBGL_INSTALL_SHARED_PATH' was not declared in this scope
   const std::string inputFilename2 = stlplus::create_filespec( stlplus::folder_append_separator( OPENMVG_SFM_WEBGL_INSTALL_SHARED_PATH ) +  inputFolder , inputName );
`

It happens for version 1.2 and HEAD of the master branch (6ce820b2844a98941a5d2543050f4e278a323800).

With kind regards,  Serge Gielkens

pmoulon:
Bonjour Serge,
@rperrot Any idea?

rperrot:
Does the issue come only with gcc 7.2 ?

sgielkens:
Hi Romuald.

Thanks for the quick response.

[...]

PR #1103, opened 2017-10-12
Patch for issue 1081:

sgielkens:
SfMWebGLViewer: compile error gcc 7.2 (Fedora 26)

The patch makes sure that during compilation /usr/include/flann/config.h
is ->not<- used but instead build/software/SfMWebGLViewer/config.h

### DIFF
--- a/src/cmakeFindModules/FindFlann.cmake
+++ b/src/cmakeFindModules/FindFlann.cmake
@@ -2,7 +2,7 @@
 #                  Find Flann Library
 #----------------------------------------------------------
 
-FIND_PATH(FLANN_DIR flann.hpp
+FIND_PATH(FLANN_DIR flann/flann.hpp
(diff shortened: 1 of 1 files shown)


src/cmakeFindModules/FindFlann.cmake
```diff
@@ -2,7 +2,7 @@
 #                  Find Flann Library
 #----------------------------------------------------------
 
-FIND_PATH(FLANN_DIR flann.hpp
+FIND_PATH(FLANN_DIR flann/flann.hpp
     HINTS "${FLANN_ROOT}" "$ENV{FLANN_ROOT}" "${FLANN_INCLUDE_DIR_HINTS}"
     PATHS "$ENV{PROGRAMFILES}/flann" "$ENV{PROGRAMW6432}/flann"
```
pmoulon: Not sure your fix will work for windows platforms...

sgielkens: No idea either, I use Linux exclusively.

sgielkens:
Proposed fix for #1081

Issue #1092, opened 2017-10-09
Build error including "sfm_data.hpp" while using openMVG as thirdparty library.


srivathsanmurali:
Hi,
I am trying to use to openMVG as a third party library.

When I include "sfm_data.hpp", it includes "sfm_view.hpp"

"sfm_view.hpp" seems to include third_party filesystem.hpp in its header which is breaking my build as third_party headers are in openMVG/third_party.
https://github.com/openMVG/openMVG/blob/6ce820b2844a98941a5d2543050f4e278a323800/src/openMVG/sfm/sfm_view.hpp#L16

Is it possible to replace "third_party" to "openMVG/third_party" when installing the files.
Hope you can help me out.

pmoulon:
Hi @srivathsanmurali,
Are you using cmake to use OpenMVG as a third party.
The include dir path must be adapted automatically here: https://github.com/openMVG/openMVG/blob/develop/src/CMakeLists.txt#L590
and so the file must be found in the include dir.

BTW, the best way to deal with this issue would be to work on using the CMAKE `INTERFACE` configuration (I need to work on this point):
https://github.com/openMVG/openMVG/blob/develop/src/third_party/lemon/lemon/CMakeLists.txt#L58

srivathsanmurali:
If I am not wrong,
https://github.com/openMVG/openMVG/blob/0813c7acf60d8c9df18be223a6707bbd43e501cd/src/CMakeLists.txt#L590-L595
this code snippet only changes the paths in the $(OpenMVG_INCLUDE_DIRS) set that have in thirdparty to appear in openMVG/third_party. It doesn't change the content of the files.

My issue is that in the sfm_view.hpp, it looks for the "file_systems.hpp".
https://github.com/openMVG/openMVG/blob/6ce820b2844a98941a5d2543050f4e278a323800/src/openMVG/sfm/sfm_view.hpp#L16
Upon on installation, it must be
``` c++
#include "openMVG/third_party/stlplus3/filesystemSimplified/file_system.hpp" 
```
A way I would go about doing it is, replace third_party to openMVG/third_party in each of the files in $(OpenMVG_INCLUDE_DIRS)

srivathsanmurali:
FYI:
I am installing openMVG in /usr/local and using find_package(openMVG) in my project.

pmoulon:
I'm not for modifying the include path in the installed files.
I will have a look soon and come back to you.

pmoulon:
[...]

PR #1105, opened 2017-10-13
fixing sfm_data.hpp include error when using openMVG as a third_party

srivathsanmurali:
Hi,
This PR request corresponds to my issue #1092 where I wasn't able to include sfm_data.hpp header because it loads a header that includes a third_party library header which can't be found.

So to fix this problem, I decided to move the inclusion of "file_system.hpp" to source files that are compiled in libraries.

I also found that many source files didn't include the header because they were available in sfm_view.hpp.
I think this could cause problems when such API files are changed.

This PR helps in cleaning up the API headers such as "sfm_data.hpp" and "sfm_data_io.hpp" which can be used without any problems.

Thanks for your time :D


pmoulon:
Thx @srivathsanmurali!

PR #1106, opened 2017-10-15
Develop - Add functionality that down samples a sfm_data.bin to a rectangular area

BigAloz:
Background,

With large datasets the dense reconstruction becomes a bottle neck in the processing due to the large amount of data involved.

Code is attached that does the following.
1. The sfm_data.bin file is assumed to be in a geodetic coordinate system
2. If it is in ECEF then a transverse mercator parameters are passed so that the conversion between ECEF and TM can be completed.
3. A down sample rectangle is passed. This define the upper NW corner and lower SE corner. This is TM coordinates.
4. Sfm_data.bin is then down sampled to include poses that only provide data to the down sample rectangle area.

Two Geodetic libraries are included. All though not implemented here it is possible to use different projections (Lambert,NZRSO,Mercater,etc). Full geodetic computations are also included which may be useful for some people.

This is my first attempt at contributing so apologise in advance for anything I get wrong in the process.
[Sfm_DownSample.zip](https://github.com/openMVG/openMVG/files/1385714/Sfm_DownSample.zip)


pmoulon:
@BigAloz Thank you for your sharing.

In order to make a contribution you must have a fork and make a PR with your modified code in it.
See here for the instructions: https://github.com/openMVG/openMVG/blob/master/CONTRIBUTING.md

whuaegeanse:
@BigAloz  Why not use the library Proj4 ?

BigAloz:
What is Proj4?

whuaegeanse:
@BigAloz Here is a link http://proj4.org/index.html

PR #1113, opened 2017-10-19
Change order or priority of intrinsics initialization (manual over exif)

hcjghr:
Hi @pmoulon!

I created a really simple fix to prioritize manually (user) provided intrinsic parameters over exif tag. My reasoning is that image can have an exif tag which provides certain estimation of intrinsics, but if we manually calibrated the camera these will be more accurate and should be used if provided. If for whatever reason manually provided parameters are not ok, the procedure will still look for exif tag.

As you will see the only thing I changed is the order of how different parts are called.

What do you think?


pmoulon:
Merged by https://github.com/openMVG/openMVG/commit/665ffc948d6848f1e43ff235456cc3673bacc426

PR #1114, opened 2017-10-20
fix building ceres with CXSPARSE=OFF

rwiesenfarth:
Setting CXSPARSE to OFF led to compile errors. I noticed that the else() part was missing in the CMakeLists.txt and added it.

I also made the formatting of the if clause similar to the other parts of the file.


pmoulon:
Thx! 👍 
I wonder which type or environment or settings are you using in order to do not use the Sparse backend ;-)
Perhaps you just want to be limited to the Eigen library for the dense and sparse computation.

pmoulon:
Thanks @rwiesenfarth 
Looking forward your next PR!

PR #1115, opened 2017-10-20
allow for external version of cereal

rwiesenfarth:
Use CEREAL_INCLUDE_DIR_HINTS to configure the include path to
an external version of cereal

I was combining OpenMVG with another library which also uses cereal, but a different version. To get the two libraries working together, I had to base them on the same version of cereal. Thus I changed OpenMVG to allow configuring an external version of cereal, similar to other third-party libraries.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -232,6 +232,20 @@ MACRO(UNIT_TEST NAMESPACE NAME EXTRA_LIBS)
 ENDMACRO(UNIT_TEST)
 
 # ==============================================================================
+# cereal
+# ==============================================================================
+# - internal by default,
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -506,6 +520,12 @@ message("** Use OpenCV SIFT features: " ${OpenMVG_USE_OCVSIFT})
 
 message("\n")
 
+if (DEFINED OpenMVG_USE_INTERNAL_CEREAL)
+  message(STATUS "CEREAL: (internal)")
```
pmoulon: While we are working on that, I wonder if we have a way to get the cereal version from the header file.
So when we are asking cmake log to our user we can diagnose more regarding the library version...

PR #1116, opened 2017-10-20
shorten names of sample programs

rwiesenfarth:
The long names led to errors about path length when built on Windows.
Shortening the names allows builds even when using non-toplevel root
directories for OpenMVG.

I have to admit that the long names of the sample executables made them both unique and descriptive about what they do. However, given CMake's method of creating file and directory names during the build process led to "path too long" errors on Windows even on rather short build directory paths.

I first tried to set CMAKE_OBJECT_PATH_MAX when configuring and building, but this did not show an effect on the file paths in question.

### DIFF
--- a/src/openMVG_Samples/cameras_undisto_Brown/CMakeLists.txt
+++ b/src/openMVG_Samples/cameras_undisto_Brown/CMakeLists.txt
@@ -1,9 +1,9 @@
 
-add_executable(openMVG_sample_cameras_undistoBrown undistoBrown.cpp)
-target_link_libraries(openMVG_sample_cameras_undistoBrown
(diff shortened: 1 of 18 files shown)


pmoulon:
Thx for your comment and PR.

I understand your concern about it.

I would like to keep it as it is for the following reasons:
- it's a platform specific issue,
- you can always build OpenMVG in a shorten path on your Windows,
- shorten the name can still make the problem appear if you try to build OpenMVG in a longer directory list. It just postpone the problem a bit,
- samples are optional at cmake configuration time.

There is certainly other way to find in order to keep the discriminative name we are using.

Lets continue to discuss about it.

Issue #1117, opened 2017-10-20
Cannot run make install with non-root user since EIGEN_INCLUDE_INSTALL_DIR has been renamed


luxigo:
Hi Pierre !

You are doing a great job, thanks again :-)

Looks like while updating Eigen something has been broken in cmake files and cmake tries to install it in the root directory.
```
CMake Error at third_party/eigen/Eigen/cmake_install.cmake:44 (file):
  file cannot create directory: /eigen3/Eigen.  Maybe need administrative
  privileges.
Call Stack (most recent call first):
  third_party/eigen/cmake_install.cmake:37 (include)
  third_party/cmake_install.cmake:65 (include)
  cmake_install.cmake:84 (include)
```
In this commit 0bad788ab3cb06b475fd3c12484e224d1ff7df3e EIGEN_INCLUDE_INSTALL_DIR has been renamed INCLUDE_INSTALL_DIR 

The real problem is probably elsewhere, but reverting to EIGEN_INCLUDE_INSTALL_DIR does the job as long backward compatibility is provided by eigen here:
https://github.com/openMVG/openMVG/blob/31c94e7ae83f7e8073b6fde915c85459bcd68414/src/third_party/eigen/CMakeLists.txt#L65

I would prefer to fix the real issue but as temporary workaround we got this.

Best regards,

Luc

pmoulon:
Can you see if I fixed this or not in the develop branch?
I made this change: https://github.com/openMVG/openMVG/commit/0bad788ab3cb06b475fd3c12484e224d1ff7df3e#diff-3ae5324d418c334c9aa264ea200ffa9fL110

luxigo:
I'm talking about the up to date develop branch.

This commit (0bad...) for which a link was included in my comment above is the commit where the renaming occurs.

Looks like somehow at install time an empty string is used instead

Probably a problem with Eigen cmake files, despite the "deprecation" message coming from them

I cannot not point it out better now.  5am, I give up.

As long as backward compatibility is provided by Eigen, it's a waste of time.

srivathsanmurali:
Hi,
I also see this bug when I try to install OpenMVG.
I think this is because Eigen doesn't handle the Install directory with backward compatibility well.
When INCLUDE_INSTALL_DIR is set in
https://github.com/openMVG/openMVG/blob/e4b686a902dc9f70bd3589e27c1026c39d721558/src/third_party/CMakeLists.txt#L108-L112
[...]

PR #1118, opened 2017-10-20
Workaround for Eigen installed in the root directory

luxigo:
temporary workaround for https://github.com/openMVG/openMVG/issues/1117

### DIFF
--- a/src/third_party/CMakeLists.txt
+++ b/src/third_party/CMakeLists.txt
@@ -107,7 +107,7 @@ add_subdirectory(fast)
 ##
 if (DEFINED OpenMVG_USE_INTERNAL_EIGEN)
   #Configure Eigen install
-  set(INCLUDE_INSTALL_DIR ${CMAKE_INSTALL_PREFIX}/include/openMVG/third_party/eigen)
+  set(EIGEN_INCLUDE_INSTALL_DIR ${CMAKE_INSTALL_PREFIX}/include/openMVG/third_party/eigen)
   add_subdirectory(eigen)
 endif()
 


pmoulon:
For me the problems comes from Eigen cmake, since they are moving people to use `INCLUDE_INSTALL_DIR` I don't understand that why if the value `INCLUDE_INSTALL_DIR` is already defined they overwrite it!, see below

```
# Backward compatibility support for EIGEN_INCLUDE_INSTALL_DIR
if(EIGEN_INCLUDE_INSTALL_DIR)
  message(WARNING "EIGEN_INCLUDE_INSTALL_DIR is deprecated. Use INCLUDE_INSTALL_DIR instead.")
endif()

if(EIGEN_INCLUDE_INSTALL_DIR AND NOT INCLUDE_INSTALL_DIR)
  set(INCLUDE_INSTALL_DIR ${EIGEN_INCLUDE_INSTALL_DIR}
      CACHE PATH "The directory relative to CMAKE_PREFIX_PATH where Eigen header files are installed")
else() # (pierre comment) if not defined EIGEN_INCLUDE_INSTALL_DIR or defined INCLUDE_INSTALL_DIR
  set(INCLUDE_INSTALL_DIR
      "${CMAKE_INSTALL_INCLUDEDIR}/eigen3"
      CACHE PATH "The directory relative to CMAKE_PREFIX_PATH where Eigen header files are installed"
      )
endif()
```

For me the test should be changed to
```
if(EIGEN_INCLUDE_INSTALL_DIR AND NOT INCLUDE_INSTALL_DIR)
  set(INCLUDE_INSTALL_DIR ${EIGEN_INCLUDE_INSTALL_DIR}
      CACHE PATH "The directory relative to CMAKE_PREFIX_PATH where Eigen header files are installed")
else()
  if (NOT INCLUDE_INSTALL_DIR)
    set(INCLUDE_INSTALL_DIR
        "${CMAKE_INSTALL_INCLUDEDIR}/eigen3"
        CACHE PATH "The directory relative to CMAKE_PREFIX_PATH where Eigen header files are installed"
        )
  endif()
endif()
```

Do you agree?

PR #1137, opened 2017-11-08
Update geodesy.rst

Yannnnnnnnnnnn:
In my test, it seems the openMVG_main_ListMatchingPairs works using sfm_data.json, instead of sfm_data.bin


pmoulon:
In fact the binary can work with any sfm_data (json, xml, bin) if the file contains some pose priors.

Yannnnnnnnnnnn:
You are right!
The sfm_data.bin and sfm_data.json are just the same for the programmer.
 But we can only get sfm_data.bin after the BA if we do not change the code,which means that there is only a sfm_data.json for us if we want to use openMVG_main_ListMatchingPairs.

Issue #1146, opened 2017-11-13
Bad MSVC preprocessor definition /bigobj


rwiesenfarth:
Compiling OpenMVG on Windows with MSVC creates a compiler command line including '/D "/bigobj"' instead of the intended '/bigobj'

Although the build succeeds, IntelliSense marks the first char in every .cpp file with an error (see attachment).

I have a working fix for this and will provide a pull request.

![openmvg_bigobj_error](https://user-images.githubusercontent.com/21051884/32732731-c4001dfc-c88d-11e7-8fb6-5f75f1b3a37e.png)

pmoulon:
I took a deeper look and in fact the /bigobj is needed only in Debug mode for the target openMVG_sfm and openMVG_matching:

I'm testing it right now there:
https://github.com/openMVG/openMVG/commit/b81a36cdb46a45525cfc3a99fad7802f81b2655b#diff-51a7011e2497130a3d10c187652e3772R46
https://github.com/openMVG/openMVG/commit/b81a36cdb46a45525cfc3a99fad7802f81b2655b#diff-c4e7d384906636d63dabada049b55348R42

pmoulon:
Can be closed since this issue is now fixed by https://github.com/openMVG/openMVG/commit/c054265a0d2ad22fd87d6bdb2f0cb276e8ac079c

PR #1147, opened 2017-11-13
fix for #1146: set /bigobj as a compiler option, not as definition

rwiesenfarth:
With MSVC, the /bigobj compiler option was added in CMakeLists.txt using register_definitions(), leading to compiler command line containing '/D "/bigobj"'. The correct command line should contain '/bigobj' instead.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -76,7 +76,7 @@ if (WIN32)
   register_definitions(-DVC_EXTRALEAN)
   register_definitions(-D_USE_MATH_DEFINES)
   if (MSVC)
-    register_definitions(/bigobj)
+    add_compile_options(/bigobj)
   endif (MSVC)
 endif (WIN32)
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -232,6 +232,20 @@ MACRO(UNIT_TEST NAMESPACE NAME EXTRA_LIBS)
 ENDMACRO(UNIT_TEST)
 
 # ==============================================================================
+# cereal
+# ==============================================================================
+# - internal by default,
+# - external if CEREAL_INCLUDE_DIR_HINTS is defined
+# ==============================================================================
+if (NOT DEFINED CEREAL_INCLUDE_DIR_HINTS)
+  set(CEREAL_INCLUDE_DIRS
+    ${CMAKE_CURRENT_SOURCE_DIR}/dependencies/cereal/include)
+  set(OpenMVG_USE_INTERNAL_CEREAL ON)
+else()
+  set(CEREAL_INCLUDE_DIRS ${CEREAL_INCLUDE_DIR_HINTS})
+endif()
+
+# ==============================================================================
 # Eigen
 # ==============================================================================
 # - internal by default,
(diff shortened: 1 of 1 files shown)


pmoulon:
BTW, I will check you update with the new Continuous Integration for VS https://github.com/openMVG/openMVG/commit/7573da84056d847f5fc44c10f9ac243b05751627

While you are working on this, can you investigate with binaries, library need this preprocessor?

pmoulon:
I have updated the CI to run VS 2015 and vs2017.
I ran a new build, let the time tell use if it compiles safely.

Did you try to delete the flag and see if the project compiles?

- we made some drastic change regarding the compilation recently and so perhaps this flag is not longer required.

rwiesenfarth:
I didn't remove the flag, so I am unable to tell if OpenMVG compiles without it. Plus, the necessity of the flag might depend on the actual configuration, so I will probably not achieve sufficient coverage, as I use a special configuration only.

I only can tell that the IntelliSense errors are gone after my change, and that the /bigobj flags appears correctly in the "Additional Options" section of MSVC's "Configuration Properties"  / "C/C++" / "Command Line".

pmoulon:
I ran again the CI on windows and curiously it fails only in one build:
https://ci.appveyor.com/project/pmoulon/openmvg/build/1.0.773
`Environment: GENERATOR=Visual Studio 14 2015 Win64, PLATFORM=x64, APPVEYOR_BUILD_WORKER_IMAGE=Visual Studio 2015; Configuration: Release`

rwiesenfarth:
That seems to be an issue of the build system loosing a (network?) connection. Looking at the lines starting here:
[https://ci.appveyor.com/project/pmoulon/openmvg/build/1.0.773/job/sf9kg8fav7rx0txk#L3409](https://ci.appveyor.com/project/pmoulon/openmvg/build/1.0.773/job/sf9kg8fav7rx0txk#L3409)

the log shows messages that are unlikely to be OpenMVG related, but indicate a problem msbuild was running into:

_C:\Tools\vcpkg\scripts\buildsystems\msbuild\vcpkg.targets(73,5): warning MSB5021: "cmd" and its child processes are being terminated in order to cancel the build. [c:\dev\openMVG\openMVG_Build\openMVG\features\openMVG_test_features.vcxproj]_

**MSBUILD : error MSB4166: Child node "Microsoft.Build.Exceptions.InternalLoggerException: The build stopped unexpectedly because of an unexpected logger failure.**

_---> System.Net.WebException: The underlying connection was closed: A connection that was expected to be kept alive was closed by the server._
_---> System.IO.IOException: Unable to read data from the transport connection: An existing connection was forcibly closed by the remote host._
_---> System.Net.Sockets.SocketException: An existing connection was forcibly closed by the remote host_

The following link refers to the same error, being related to AppVeyor:

[https://help.appveyor.com/discussions/problems/890-appveyorlogger-causing-microsoftbuildexceptionsinternalloggerexception](https://help.appveyor.com/discussions/problems/890-appveyorlogger-causing-microsoftbuildexceptionsinternalloggerexception)

bjornpiltz:
I apprieciate fixing this nuisance, but I think a different approach is necessary. The problem is actually [this block](https://github.com/openMVG/openMVG/blob/6ce820b2844a98941a5d2543050f4e278a323800/src/CMakeLists.txt#L611-L613):
```cmake
foreach(d ${OpenMVG_DEFINITIONS})
  list(APPEND OpenMVG_CFLAGS "-D${d}")
endforeach()
```
which prepends `-D` to `/bigobj`. Most likely /bigobj isn't needed anyway, since it isn't applied at the moment afaict. So I suggest either:

- Don't prepend `-D` to `/bigobj`, or
- don't use  `/bigobj` anymore.

pmoulon:
This problem will be fixed soon in the branch develop_cmake. See https://github.com/openMVG/openMVG/commit/027ba010e20e98e24486c5f1b108febc4687f06f#diff-4d3c97ab21894254d5eb10604ece6864R32

pmoulon:
`develop_cmake` is now merged to develop. Problem solved ;-)

Thx!

PR #1151, opened 2017-11-20
Fix link in README.md

timmd909:
Pretty self explanatory...


pmoulon:
Thanks @timmd909 We already fixed this issue in the develop branch and it will be pushed to master soon with OpenMVG v1.3.

PR #1185, opened 2017-12-20
remove useless argument passing

sebastienchappuis:
Hello,
This is a tiny PR to remove the `SfM_Data` argument in the `Matcher::Match(...)` method (and derived classes equivalents). It is never actually used in the matcher.

Cheers

### DIFF
--- a/src/openMVG/matching_image_collection/Cascade_Hashing_Matcher_Regions.cpp
+++ b/src/openMVG/matching_image_collection/Cascade_Hashing_Matcher_Regions.cpp
@@ -36,7 +36,6 @@ namespace impl
 template <typename ScalarT>
 void Match
 (
-  const sfm::SfM_Data & sfm_data,
   const sfm::Regions_Provider & regions_provider,
   const Pair_Set & pairs,
   float fDistRatio,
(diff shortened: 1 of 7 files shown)


pmoulon:
Thank you for your comment and PR.
It is true that for the moment we don't have any matcher that use the information contained in the sfm_data. I thought that it would be cool to pass along the sfm_data in order to relate the pair ID to image, camera or existing pose information that could exist if a given matching method would need it.

I think that for the moment it is safer to use as less as possible interface and bring it back later on if we need it ;-)

whuaegeanse:
@pmoulon @sebastienchappuis If the intrinsics of camers in the sfm_data are known, it would be more robust to removing the distortion before matching the image pairs.

pmoulon:
@whuaegeanse As you can see here, if the camera model has some existing distortion point are already undistorted for geometric model estimation ;-)

In OpenMVG, there is the putative matches step (that just use the descriptor) and then the geometric filtering (that perform undistortion if necessary).

sebastienchappuis:
Ah ok, I thought this was a remnant of some older design that got forgotten in here :)
I think it's better to remove it, since it can lead to some cases where you have to create an empty sfm_data object to be able to use the matcher...which is confusing.

not sure why there is a segfault in one of the windows build though...

pmoulon:
@sebastienchappuis The problem with the CI and VS2017 is a compiler issue.
@rperrot made an investigation and found that the issue come from this area:
`observation(distribution(generator)) = pertubations(i);`
In fact @rperrot  found that it is related to this https://developercommunity.visualstudio.com/content/problem/171811/msvc-1552-incorrect-assembly-generated-when-optimi.html

The VS team already made a fix but we have to wait the next version.

pmoulon:
Thx!

Merged by https://github.com/openMVG/openMVG/commit/d815f7d26befb302507984061a80f663f0c53d6e

PR #1199, opened 2018-01-09
Fix typo in ply export code

aiqu:
Export control points in red.
Currently, it exports control points in white if save with binary format.


pmoulon:
Thx merged by https://github.com/openMVG/openMVG/commit/63d1b275065095da6d44aa4bbd7285964226a794

PR #1201, opened 2018-01-10
openMVG2openMVS - Parallelization of export of undistorted images

hcjghr:
Hi,

this PR adds the ability of parallel export of undistorted images in the process of exporting data to openMVS. To parallelise it, I moved parts of the code (relating the undistortion of images) from the "define images & poses" for-loop in exportToOpenMVS to a new for-loop. 

Cheers,
Klemen

### DIFF
--- a/src/software/SfM/main_openMVG2openMVS.cpp
+++ b/src/software/SfM/main_openMVG2openMVS.cpp
@@ -35,7 +35,8 @@ using namespace openMVG::sfm;
 bool exportToOpenMVS(
   const SfM_Data & sfm_data,
   const std::string & sOutFile,
-  const std::string & sOutDir
+  const std::string & sOutDir,
+  const int iNumThreads = 0
   )
 {
   // Create undistorted images directory structure
@@ -96,7 +97,7 @@ bool exportToOpenMVS(
       std::cout << "Cannot read the corresponding image: " << srcImage << std::endl;
       return EXIT_FAILURE;
     }
-    if (sfm_data.IsPoseAndIntrinsicDefined(view.second.get()) && stlplus::is_file(srcImage))
+    if (sfm_data.IsPoseAndIntrinsicDefined(view.second.get()))
     {
       MVS::Interface::Platform::Pose pose;
       image.poseID = platform.poses.size();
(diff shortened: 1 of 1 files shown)


pmoulon:
Thanks @hcjghr ,
For the moment I did not made this step in parallel since depending of your configuration it can leads to crash since the process of undistorting image can take more RAM than your pc have.

I have done the parallelization here:
https://github.com/openMVG/openMVG/blob/master/src/software/SfM/main_ExportUndistortedImages.cpp#L98

But I think in order to avoid a segfault and so have a user that does not understand what's going on. We must use somehow at least a `try catch` to catch an eventual segmentation fault due to memory allocation reason. This way we can tell to the user that the process crash since we are using too much ressources and guide him to use less threads.

What do you think about that?

The other point is that this image undistortion step is required by many exporter and so perhaps we should build a mini interface with functor to generate the view output name.
So this way we could call the same interface for all the exporter and only customize the function that generate the view output filename.

hcjghr:
@pmoulon you make some good points.

The first thing I want to check with you is: 

Is there currently a way to avoid exporting (undistorting images) in the openMVG2openMVS binary? If not, then even if a user has multiple cores and sufficient enough RAM he is stuck with slow exportation of images (and another binary doesnt help at all) - which is in my opinion a really bad solution and poses a real bottleneck.

Addressing the memory allocation problems, I think the try/catch solution together with the default setting of using 1 thread would be probably the most elegant solution. This way, the user who does not intentionally set (i.e. -n) the use parallel export  will not unknowingly run into problems, however it gives users that wish to an easy way to exploit the hardware.

Regarding the unified image undistortion step, I completely agree that the code structure will benefit from it.

pmoulon:
OpenMVS uses undistorted images. So there is no way to avoid the image undistortion.
The only smart thing we can do is that if the camera model is the classic pinhole (-c 1), so the camera does not have a distortion field and so we can just copy the image.

Let's work on finding a way to factore the code that handle the undistortion and handle in a nice way the (memory allocation trouble -> try catch) and the parallelization (omp loop).

pmoulon:
Thx @hcjghr PR updated (for multithread allocation error checking) and merged

PR #1213, opened 2018-01-16
Partially replace OpenMP with TBB

tomas789:
tldr: I propose to replace OpenMP with TBB. This is mostly draft to open a discussion.

OpenMVG currently supports only OpenMP for parallelisation. I couldn't get it working on OS X. Problem is that it is not supported by version of clang shipped with Xcode (majority of OS X people have only this one). There are [some options](https://stackoverflow.com/questions/35134681/installing-openmp-on-mac-os-x-10-11) but it is fare from being flawless. 
 - You need to instal another compiler.
 - Widely recommended is GCC but if you have some custom-compiled libraries, you have to recompile them (clang built binaries don't have same ABI as GCC compiled ones)

This very much limits usage of OpenMVG on OS X (and AFAIK similar situation is on Windows). Nowadays, it is very easy to get multithreading work. I would like to open discussion on replacing OpenMP with something easier to adopt.

Possible replacements:
*Platform specific threading*: I guess not really an option. Too difficult to develop, even more difficult to maintain.
*C++11 threads*: Very simple to adopt, does not have support for higher level primitives (such as parallel_for).
*Thread Building Blocks*: As simple to adopt as C++11, supports all primitives needed, provides big opportunities for future parallelisation.

Based on this, I implemented sketch of how can we implement TBB into OpenMVG and replace OpenMP. 

Comments:
 - I made TBB mandatory (is there a reason not to?).
 - We can use TBB in built in ceres-solver to guarantee access to multithreading there.
 - I did not port all parts to TBB. This is meant to be a sample.
 - It shouldn't take more then few hours to port everything (mostly because TBB's parallel_for is super compatible with it's counterpart from OpenMP).

I would like to see OpenMVG being able to run easily and blazing fast on OS X.


pmoulon:
Hi @tomas789 

Appreciated your PR with comments and proof of concept!

- I agree that parallelism should be a point that we can discuss and I appreciate your concern for the  library usability on many platform in a uniform way!

As you said, parallelization/multithreading can be done today in various way:
- openMP
- C++11 threading (used only in some OpenMVG module)
- TBB
- Cilk Plus (no longer maintain - deprecated)
- ...

More or all the solutions have pros and cons.

Since the main objective of OpenMVG is to bring knowledge about computer vision and multiple view geometry to the community, the community focused on an option that has a minimal impact on the requirement (the compiler bring support or not of the technology). I mean OpenMVG can compile with or without OpenMP.

Somehow we can say that the user is limited by:
- the dev platform:
  - in your case (MacOS you are blocked due to your compiler, not the library.)
- by the used technology:
  - OpenMP is not supported on Android, and so using another framework we can can extend the access or parallel code to other platform ;-).

Note that we have used OpenMP since it is:
- easy to use (minimal modification of the code)
- supported by default by a lot of compiler (I still does not get why MacOS Mac does not support is by default)
- rely on an "open" standard organization and not a single organization (Not Intel, Not Microsoft, ...)

The good point about C++11 threading and TBB is that they rely on C++ specifications and so rely on high level object.

Somehow we know that if we want to switch to C++11 or TBB it will involve more lambda code and so change some code from
`for {}` to `for( [](...) {} );`

I'm totally open to give a try for TBB but I would like some answer to those questions on TBB and some comment from the community on your post ;-)

- Do we can customize the number of thread (by an internal function or a env variable)?
- Do we can manage to make TBB only an optional dependency?
- How TBB manage a parallel loop inside a parallel loop (nested vs not nested)?
- Does the TBB parallel_for is working on iterator or c++11 range?

As a short term action plan we can do this:
- Try to gather the community and make everyone happy to discuss about this @rperrot @bjornpiltz @rjanvier...

- Give a try to this for a module (i.e. feature extraction):
  - implement the solution, the CI compilation (Travis, AppVeyor)
  - look at the code and performance and decide which solution we will use.

Thanks you @tomas789, I'm impatient to continue the discussion.

tomas789:
Thank you for reaction @pmoulon. Here is a quick shot to answer your questions:

- **Do we can customize the number of thread (by an internal function or a env variable)?**
   This can be done in TBB. In fact, [I'm already doing this](https://github.com/openMVG/openMVG/pull/1213/files#diff-31e8d4fb05c8c87d50a2bf57a2f3e191R125) to support `-n` on command line arguments. It facilitates [`tbb::task_scheduler_init`](https://software.intel.com/en-us/node/506296). You can either let TBB decide how many threads to use (used when you specify `-n 0` or no `-n` at all; defaults to hardware concurrency value, similar to [`std::thread::hardware_concurrency`](http://en.cppreference.com/w/cpp/thread/thread/hardware_concurrency))
- **Do we can manage to make TBB only an optional dependency?**
   It probably can be done. Easy way to achieve this would be to provide custom wrapper around `tbb::parallel_for` and `tbb::parallel_for_each` with the same interface and provide two implementations. Question is whether this effort makes OpenMVG better. Two reasons. First, TBB is relatively small dependency that works multiplatform (including iOS and Android), so it should not be a big problem to require it (can be shipped internally). Second, for most use cases, threads are mandatory. Single thread performance is poor even for moderate-size use cases. On the other hand, the way I'm using this is very specific so don't take me seriously here :)
- **How TBB manage a parallel loop inside a parallel loop (nested vs not nested)?**
    This is handled properly. It is already used in PR [here](https://github.com/openMVG/openMVG/pull/1213/files#diff-3df0166a53a1a99043a688f8af017f9dR130). In TBB, when you use `parallel_for`, every iteration of loop (not totally true, it generates ranges that are being split so even multiple iterations of for-loop can be done as single task by one thread) is scheduled and executed on thread pool. Inner `parallel_for` then becomes just another task in the queue.
- **Does the TBB parallel_for is working on iterator or c++11 range?**
  Yes, using `tbb::parallel_for_each`. Example is [here](https://github.com/openMVG/openMVG/pull/1213/files#diff-3df0166a53a1a99043a688f8af017f9dR130).

donlk:
Quick clarification:
On Android proper OpenMP threading support is down to the used libc library and the compiler itself, not the OS. GCC and Clang both have quite mature OMP support for quite some time now you can build such a cross toolchain yourself to build with or just use the built-in ones from the current NDK package. Either way, proper OMP support for Android is a solved problem.

rperrot:
Thanks @tomas789 for this PR. 

Regarding runtimes performances are there any differences between openMP, TBB and (if somebody has stats) c++11 threads ?

I think the more portable solution is to use c++11 std::thread but it's verbose and longer to implement (almost everything needs to be defined by hand). 

To all : do you know some alternatives to theses codes : 
http://www.alecjacobson.com/weblog/?p=4544
http://dlib.net/parallel_for_ex.cpp.html

that: 
- would be portable 
- would use omp/tbb/c++11 depending on the available library/tools
- would be well written
- would produce easy readable code 
- would be type safe
- won't pollute debug mode. 

? 

Maybe we can think a little higher level for a solution that can be acceptable to everybody.

rhiestan:
A small note from my side: Although the Clang compiler shipped with XCode does not support OpenMP, it is very easy to install the clang compiler from the LLVM homepage (http://clang.llvm.org/). This one does support OpenMP, and (as far as I can tell) is binary-compatible with the XCode-Clang. I compile OpenMVG and Regard3D using the Clang from the LLVM homepage.

RamadanAhmed:
I think we can use C++17 parallel algorithms library although currently only few c++ compiler support it(MSVC,Intel), but in near future it will be easy to remove openmp.

pmoulon:
Closing PR since the project prefer rely on none TBB threading

PR #1220, opened 2018-01-25
Fixed typo in binary ply generation. GCP should be red

MatillaMartin:


MatillaMartin:
Closing. Already fixed in develop

PR #1235, opened 2018-02-20
cmake improvements

mitjap:
Main idea behind this PR is to enable lib users to use modern cmake to link to OpenMVG

For example in my demo project this looks like this
```

find_package(OpenMVG REQUIRED)
...
target_link_libraries(survey_algorithms_lib
    PRIVATE
        Qt5::Core
        OpenMVG::openMVG_sfm
        Eigen3::Eigen
        Boost::boost
    PUBLIC
        survey::core
)
```

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -127,6 +127,16 @@ $ git clone --recursive https://github.com/openMVG/openMVG.git
 6. Change the target to Release.
 7. Compile the libraries and binaries samples.
 
+Another options is to build with cmake in console. I recommend launching `VS2015 x64 Native Tools Command Prompt` which has build environment already set up.
+
+```shell
+git clone --recursive https://github.com/openMVG/openMVG.git
+mkdir openMVG-build
+cd openMVG-build
+cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX="c:/openMVG" -G "NMake Makefiles" ../openMVG/src/
+cmake --build . --target install
+```
+
 Mac compilation
 -------------------
 <a name="macos"></a>
@@ -182,6 +192,18 @@ find_package(OpenMVG REQUIRED)
 include_directories(${OPENMVG_INCLUDE_DIRS})
 add_executable(main main.cpp)
 target_link_libraries(main ${OPENMVG_LIBRARIES})
+```
+
+or with modern target-based approch (CMake 3.0+)
+
+```
+find_package(OpenMVG REQUIRED)
+add_executable(main main.cpp)
+target_link_libraries(main
+  PRIVATE
+    OpenMVG::openMVG_sfm
+    OpenMVG::openMVG_matching
+)
 ```
 
 Specify in CMake where OpenMVG has been installed by using the CMake OpenMVG_DIR variable
--- a/BUILD.md
+++ b/BUILD.md
@@ -194,7 +194,7 @@ add_executable(main main.cpp)
 target_link_libraries(main ${OPENMVG_LIBRARIES})
 ```
 
-or with modern target-based approch (CMake 3.0+)
+or with modern target-based approach (CMake 3.0+)
 
 ```
 find_package(OpenMVG REQUIRED)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -579,7 +579,10 @@ message("\n")
 
 install(EXPORT openMVG-targets DESTINATION lib)
 install(EXPORT openMVG-targets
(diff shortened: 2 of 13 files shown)


BUILD.md
```diff
@@ -184,6 +194,18 @@ add_executable(main main.cpp)
 target_link_libraries(main ${OPENMVG_LIBRARIES})
 ```
 
+or with modern target-based approch (CMake 3.0+)
```
pmoulon: typo `approch ` -> `approach`

src/CMakeLists.txt
```diff
@@ -579,7 +579,10 @@ message("\n")
 
 install(EXPORT openMVG-targets DESTINATION lib)
 install(EXPORT openMVG-targets
-        DESTINATION share/openMVG/cmake FILE OpenMVGTargets.cmake)
+  FILE OpenMVGTargets.cmake
+  NAMESPACE OpenMVG::
```
pmoulon: Does the `::` are necessary?

src/openMVG/features/CMakeLists.txt
```diff
@@ -24,7 +24,8 @@ target_include_directories(openMVG_features
     $<BUILD_INTERFACE:${EIGEN_INCLUDE_DIRS}>
     $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}>
     $<BUILD_INTERFACE:${CEREAL_INCLUDE_DIRS}>
-    #$<INSTALL_INTERFACE:include/openMVG/>
+    $<INSTALL_INTERFACE:include/>
```
pmoulon: Seems like we can remove the last `/`

src/openMVG/matching/CMakeLists.txt
```diff
@@ -24,14 +24,23 @@ add_library(openMVG_matching
   ${matching_files_header}
   ${matching_files_cpp})
 target_include_directories(openMVG_matching
-  PUBLIC $<BUILD_INTERFACE:${FLANN_INCLUDE_DIRS}>
+  PUBLIC
+    $<BUILD_INTERFACE:${FLANN_INCLUDE_DIRS}>
+    $<INSTALL_INTERFACE:include/>
```
pmoulon: We need to stick to `/` or does not put it at the end.
Seems like cmake recommend not to use the last `/`.
See at the end of https://cmake.org/cmake/help/v3.0/command/install.html

src/CMakeLists.txt
mitjap: Yes, interestingly it is necessary.

mitjap:
Thank you for review. I'll change this first thing tomorrow. I'm also rewriting our application to use modern CMake approach so if I learn enough I might be able to further improve CMake here and make it modular.

It would be nice if it was possible to use it this way.
`find_package(OpenMVG 1.3 COMPONENTS multiview sfm image geodesy REQUIRED)`

Another nice feature is to allow multi-type package installations. That would install both debug and release libraries and binaries to install path.

mitjap:
While at it I also noticed that Eigen gets installed to `/` directory on linux or `C:/Program Files` on windows. On second install it gets installed to correct directory. After some investigation I found that this has already been fixed and later reintroduced while you attempted to avoid using deprecated `EIGEN_INCLUDE_INSTALL_DIR`. On master ([here](https://github.com/openMVG/openMVG/blame/master/src/third_party/CMakeLists.txt#L110)) still works, while on develop does not ([here](https://github.com/openMVG/openMVG/blob/a53acb89e85a7a88bc541b32895dd4e7cf2bb0c3/src/third_party/eigen/CMakeLists.txt#L74)).

pmoulon:
The issue for Eigen is inside the Eigen cmake install code. They do not overwrite the right value. So depsite a warning is displayed they are not doing the right job.

pmoulon:
> Unstable AppVeyor CI

I feel like the free instances are less reliable than the pro ones and so sometimes it fails randomly.

> Eigen issue

We have to track the update on their side

pmoulon:
Do you have any idea to support this `find_package(OpenMVG 1.3 COMPONENTS multiview sfm image geodesy REQUIRED)` one day?

pmoulon:
Thx. Merged by https://github.com/openMVG/openMVG/commit/07ca8b26a828e51bb853abb0d6b8d1ab8b0b5946

mitjap:
I do. I have done this with one of my library.

pmoulon:
Looking forward to see if we can do that for OpenMVG one day ;-)

PR #1236, opened 2018-02-20
main_openMVG2Colmap executable 

SBCV:
This PR adds a main_openMVG2Colmap executable, which converts OpenMVG's SfM output to Colmap's SfM format and allows 

- to apply Colmaps MVS pipeline using OpenMVG SfM reconstructions
- to visualize reconstructions of OpenMVG using Colmaps GUI (e.g. which 3D points are visible in which frame)

Currently, one can use main_openMVG2PMVS to export data from OpenMVG to Colmap in order to use Colmap's MVS pipeline. 
However, this has one big drawback: openMVG2PMVS uses the Bundler file format, which does not contain principal point information (i.e. Colmap treats the imported SfM result, as if the principal point has been in the image center )

To create Colmap's MVS input one has to apply 
`./colmap image_undistorter`
to the output generated by 
`main_openMVG2Colmap`
To create Colmap's MVS output one has to apply (as with native Colmap SfM reconstructions)
 `./colmap dense_stereo` and `./colmap dense_fuser` 

If you have any questions or change requests, let me know.

Cheers
Sebastian


pmoulon:
Hi @SBCV 
Thank you for your PR!

- I will try to make a review during the week end.

- When I made a fast reading, I saw that the default camera model of OpenMVG `PINHOLE_CAMERA_RADIAL3 ` is not supported. Perhaps we should display a warning there to explain the user that if they want to use this MVS back-end they have to change the camera model type and rerun a BA.

For completeness we can think about listing how to use the exporter there (documentation): https://github.com/openMVG/openMVG/tree/master/docs/sphinx/rst/software/MVS

- Do you mind to share some results (screenshot or point cloud) you obtained?
  - It would be interesting to see on this dataset the output https://github.com/openMVG/ImageDataset_SceauxCastle

mitjap:
Uploded the Sceaux Castle to sketch fab - https://skfb.ly/6wNAD

mitjap:
Since I can not make a PR to a PR I'll just post it here.

```
    case PINHOLE_CAMERA_RADIAL3: 
      //OpenMVG's PINHOLE_CAMERA_RADIAL3 corresponds to Colmap's FULL_OPENCV
      //Parameters: fx, fy, cx, cy, k1, k2, p1, p2, k3, k4, k5, k6
      {
        std::shared_ptr<openMVG::cameras::Pinhole_Intrinsic_Radial_K3> pinhole_intrinsic_radial(
            dynamic_cast<openMVG::cameras::Pinhole_Intrinsic_Radial_K3 * >(intrinsic->clone()));
        
        came_line_ss << camera_id << " " << 
          "FULL_OPENCV" << " " <<
          pinhole_intrinsic_radial->w() << " " << 
          pinhole_intrinsic_radial->h() << " " <<
          pinhole_intrinsic_radial->focal() << " " << 
          pinhole_intrinsic_radial->focal() << " " << 
          pinhole_intrinsic_radial->principal_point().x() << " " << 
          pinhole_intrinsic_radial->principal_point().y() << " " << 
          pinhole_intrinsic_radial->getParams().at(3) << " " << //k1
          pinhole_intrinsic_radial->getParams().at(4) << " " << //k2
          0.0 << " " << //p1
          0.0 << " " << //p2
          pinhole_intrinsic_radial->getParams().at(5) << " " << //k3
          0.0 << " " << // k4
          0.0 << " " << // k5
          0.0 << "\n";  // k6
      }
      break;
```

This is OpenMVG Radial_3 camera converted to Colmap's FULL_OPEN_CV camera model.

SBCV:
@mitjap 
Thanks! You are right, we can just convert PINHOLE_CAMERA_RADIAL3 to  FULL_OPENCV

@pmoulon 
I added a short usage description to the documentation.
Do you need something else?

src/software/SfM/main_openMVG2Colmap.cpp
```diff
@@ -0,0 +1,436 @@
...
+    {
+      throw std::string( "Invalid command line parameter." );
+    }
+    cmd.process( argc, argv );
+  }
+  catch ( const std::string& s )
+  {
+    std::cerr << "Usage: " << argv[0] << '\n'
+              << "[-i|--sfmdata] filename, the SfM_Data file to convert\n"
+              << "[-o|--outdir] path where the cameras.bin, images.bin and points3D.bin will be saved\n"
```
mitjap: it creates `txt` files not `bin`

SBCV: Fixed

TheDoubleB:
I believe this line could be removed as `iterIntrinsic` variable remains unused.
https://github.com/openMVG/openMVG/pull/1236/files#diff-b6f61bcf3a45f30b6acf8d201e9e391aR213

SBCV:
@TheDoubleB  Thanks! - fixed

SBCV:
When I started writing this PR, I followed the structure of main_openMVG2NVM.cpp. 
In main_openMVG2NVM.cpp the color of the points is set to a constant value:
`file << 250 << " " << 100 << " " << 150 << " ";  // Write arbitrary RGB color`
This is why I did it in the same way - however, this may be annoying when visualizing the sparse reconstruction with Colmap (in the MVS case it doesn't matter) 

In `main_ComputeSfM_DataColor.cpp` there is a function `ColorizeTracks` to determine the colors of landmarks. Since `main_ComputeSfM_DataColor.cpp`has no corresponding header file I can't access this function and to avoid redundancy I do not really want to copy this function. Is there maybe some other function (already defined somewhere in openMVG), which I can exploit to export the real color of the 3D points?

pmoulon:
- We can think about adding a `sfm_data_colorization.hpp/cpp` in openMVG/sfm with this `ColorizeTracks` function.

SBCV:
This is probably the cleanest solution. Do you want me to create a separate PR? (For moving the colorization?)

pmoulon:
Yes, please do a separate PR so this way we can ensure we are using the new function in the Colorize binary ;-) and iterate if we need to do some update there.

SBCV:
@pmoulon one quick question: I just tried to add the sfm_data_colorization.cpp / .hpp to openmvg (inside src/openMVG/sfm). However, when compiling the code I run in a linker error.
` undefined reference to `openMVG::sfm::SomeFunction()`
Do I have to add sfm_data_colorization.cpp / .hpp to a specific CMake file?
Adding the function to an existing file works without any issues

pmoulon:
If you add a cpp file just launch again cmake. Since I'm using glob it does not detect the new file. (you can also fake edit the cmake so make will rerun).

SBCV:
Here is an example image showing an openmvg reconstruction in colmap. The pink points are visible in the pink camera (manually in colmap selected)  

openmvg_to_colmap

src/software/SfM/main_openMVG2Colmap.cpp
```diff
@@ -0,0 +1,447 @@
...
+  {
+    std::cerr << std::endl
+              << "The input SfM_Data file \"" << sSfM_Data_Filename << "\" cannot be read." << std::endl;
+    return EXIT_FAILURE;
+  }
+
+  if ( ! exportToColmap( sfm_data , sOutDir ) )
+  {
+    std::cerr << "There was an error during export of the file" << std::endl;
+    exit( EXIT_FAILURE );
```
pmoulon: `return EXIT_FAILURE;`

SBCV: Fixed

SBCV: Fixed

pmoulon:
@SBCV Can you explain how you open the results colmap files within the GUI?

pmoulon:
Found it `File Import model` ;-)

pmoulon:
Thx. PR merged thx to https://github.com/openMVG/openMVG/commit/53fd94aca0cda61669def36a3bae3e415b10e0d6 (Your squashed commit + some minor fix https://github.com/openMVG/openMVG/commit/07c6209e33809a17a9df5249671d63bca6412248 )

SBCV:
Thx :)

mitjap:
Awesome. One option to support all OpenMVG's camera types is to undistort
images and export to colmap as simple pinhole camera.

PR #1238, opened 2018-02-21
Ability to preserve intrinsics in main_ComputeStructureFromKnownPoses

edanvoye:
This PR adds a command line flag to main_ComputeStructureFromKnownPoses to be able to run Bundle Adjustment, but preserve the Intrinsics. In that case, the bundle adjustment only updates Extrinsics and Structure.

A typical use case would be a workflow where we want to reuse intrinsics (because we are using the same cameras and focus) but we want to be robust against any slight movement of those cameras. In this case, we want to compute intrinsics once, and reuse the poses from shot to shot to preserve general alignment and scene scale, but the bundle adjustment on extrinsics and structure will repair any camera vibration or small movement.

I used a new switch -e for this behavior because I did not want to break the existing interface. But a better command line interface could be a single switch for the bundle adjustment with flags for intrinsics, extrinsics and structure.


hwang051785:
Hi, any update on this PR? This is also a cool feature that will contribute to our current research. Thanks.

PR #1250, opened 2018-03-08
Modify CMakeLists to use openMVG on other project

aiqu:
1. Fix stlplus3 CMakeLists to generate proper include paths
2. Add & export dependencies/cereal to openMVG-targets
3. Add & export header only third_party folders to openMVG-targets

While I'm using openMVG in my project, I suffered a lot of troubles especially related to include paths.
This pr mainly focused on proper include path configuration.

Also, it should update dependencies/cereal submodule, too.
So, I'll create the corresponding pr to that repo.

Since I just learned about modern CMake, let me know if I misunderstood your intention.


aiqu:
This pr requires below.

Related pr on cereal.
https://github.com/openMVG-thirdparty/cereal/pull/1

src/openMVG/cameras/CMakeLists.txt
```diff
@@ -1,12 +1,10 @@
...
 add_library(openMVG_camera INTERFACE)
-target_link_libraries(openMVG_camera INTERFACE openMVG_numeric openMVG_geometry)
+target_link_libraries(openMVG_camera INTERFACE openMVG_numeric openMVG_geometry cereal)
 install(TARGETS openMVG_camera DESTINATION lib EXPORT openMVG-targets)
 
 add_library(openMVG_camera_test INTERFACE)
-target_link_libraries(openMVG_camera_test INTERFACE openMVG_camera openMVG_multiview)
-target_include_directories(openMVG_camera_test
-  INTERFACE $<BUILD_INTERFACE:${CEREAL_INCLUDE_DIRS}>)
+target_link_libraries(openMVG_camera_test INTERFACE openMVG_camera openMVG_multiview cereal)
```
pmoulon: ceresl is already linked in openMVG_camera so you do need to add it there.

src/third_party/CMakeLists.txt
```diff
@@ -119,4 +119,10 @@ install(
   COMPONENT headers
   FILES_MATCHING PATTERN "*.hpp" PATTERN "*.h"
 )
+add_library(${inDirectory} INTERFACE)
```
pmoulon: Can you explain why you need this?

src/CMakeLists.txt
```diff
@@ -377,6 +377,10 @@ if (OpenMVG_USE_OPENCV)
   endif (NOT OpenCV_FOUND)
 endif (OpenMVG_USE_OPENCV)
 
+# ==============================================================================
+# Dependency libraries:
+# ==============================================================================
+add_subdirectory(dependencies/cereal)
```
pmoulon: We need to do this only if we use the internal cereal. Please note that OpenMVG can use an external cereal library.
Can you check that your modification will work even you are using an external cereal library.

pmoulon:
Thx you.
Can you give you a typical main.cpp and CMakeLists.txt that you are using to check and show that the existing CMake code was not working for your purpose.

pmoulon:
FYI, your changes are not working on VisualStudio CI: https://ci.appveyor.com/project/pmoulon/openmvg/build/1.0.861

src/openMVG/cameras/CMakeLists.txt
aiqu: Got it, I'll remove cereal from openMVG_camera_test

src/third_party/CMakeLists.txt
aiqu: At first, I want to declare every third-party utility explicitly so it gives more modularity.
However, just adding 'include/openMVG' to OPENMVG_INCLUDE_DIRS would be enough.

src/CMakeLists.txt
aiqu: Got it.

aiqu:
### Example
- CMakeLists.txt
```
CMAKE_MINIMUM_REQUIRED(VERSION 3.0)

set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")

find_package(OpenMVG REQUIRED)
find_package(Eigen3 REQUIRED)

add_executable(main main.cpp)
target_include_directories(main PUBLIC
  ${OPENMVG_INCLUDE_DIRS}
  ${EIGEN3_INCLUDE_DIRS}
)
target_link_libraries(main PUBLIC
  ${OPENMVG_LIBRARIES}
)
```
- main.cpp
```
// To use cereal input/output function
#include <cereal/archives/json.hpp>
// No use case in this example, but requires in particular cases.
// Moreover, since this header is exposed to user, I think it should be included properly
#include <openMVG/graph/graph.hpp>
// It uses progress utility.
#include <openMVG/sfm/pipelines/sfm_features_provider.hpp>
#include <openMVG/features/image_describer.hpp>
#include <memory>

int main() {
  std::ifstream file("test.json");
  std::shared_ptr<openMVG::features::Image_describer> desc;
  cereal::JSONInputArchive archive(file);

  return 0;
}
```
- Encountered errors on branch develop
  1. fatal error: cereal/archives/json.hpp: No such file or directory
  2. fatal error: lemon/connectivity.h: No such file or directory
  3. fatal error: third_party/progress/progress_display.hpp: No such file or directory

pmoulon:
I will look to your PR.

FYI: According the last change those lines should not be required:

```
set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
find_package(OpenMVG REQUIRED)
find_package(Eigen3 REQUIRED)
target_include_directories(main PUBLIC
  ${OPENMVG_INCLUDE_DIRS}
  ${EIGEN3_INCLUDE_DIRS}
)
```
All the includes and compilation flags should be added by:
```
target_link_libraries(main PUBLIC
  ${OPENMVG_LIBRARIES}
)
```

PR #1253, opened 2018-03-14
FIx the CVPR 2017 resection solver

rjanvier:
Num samples where copy pasted from the 6 point DLT solver file. This could lead to some bad behavior into the ACRansac framework.


pmoulon:
Thx, merged by https://github.com/openMVG/openMVG/commit/c34171e64a79dbf948fd3f414625a259f6a0a824

PR #1254, opened 2018-03-15
[Software] Remove useless code in IncrementalSfM

rjanvier:
at this step the triangulation is useless since we do use the landmarks for BA yet.
This will hopefully speed up the computation (a bit)


pmoulon:
Thx, merged by https://github.com/openMVG/openMVG/commit/d37236d78dac7940b03a91471466e29c037ffa46

PR #1256, opened 2018-03-16
Removal of redundant ColorizeTracks function

SBCV:
The ColorizeTracks() function in openMVG_main_openMVG2WebGL.cpp is identical to the ColorizeTracks() function previously defined in main_ComputeSfM_DataColor.cpp


pmoulon:
Thx!

PR #1266, opened 2018-03-28
Fix for use of external JPEG, PNG, and TIFF libraries

TheDoubleB:
`openMVG_image` module does not compile if one is using external JPEG, PNG, and/or TIFF libraries due to missing include directories. This PR fixes the problem.


pmoulon:
Merged by https://github.com/openMVG/openMVG/commit/262d88eb6328ba99ae736aca656955fac3a03474

PR #1269, opened 2018-03-29
SfM regions provider include in ACRobust header files fix

TheDoubleB:
All three `*_ACRobust.hpp` files are missing SfM regions provider includes needed by `Regions_Provider::get()` calls in these files. This PR fixes the problem.


pmoulon:
Merged by https://github.com/openMVG/openMVG/commit/92c1b5fdf4bbbd71198acd9a25199c741b455c70

Issue #1265, opened 2018-03-28
FindFlann.cmake missing linker flag -llz4.


bartoszek:
I stumble on a problem with [src/cmakeFindModules/FindFlann.cmake](https://github.com/openMVG/openMVG/blob/db92617fea420363abcbad1cb3773706f75012ad/src/cmakeFindModules/FindFlann.cmake)
It uses custom, cmake script, where [openMVG/src/third_party/flann/cmake/FindFlann.cmake](https://github.com/openMVG/openMVG/blob/db92617fea420363abcbad1cb3773706f75012ad/src/third_party/flann/cmake/FindFlann.cmake) uses `pkg-config`.
On my system, `flann` requires `-llz4` flag (flann:v1.9.1) which gets included when using `pkg-config` based `FindFlann.cmake` (the `FLANN_DEFINITIONS` variable).
Could we combine those two, first import `pkg-config` and check if it finds `flann` and if not resolve to `find_path()`
_refers to: openmvg#branch=develop_

_Also `fast` is missing from `target_link_libraries` in `software/VO`_
_Also `lemon` insted of `emon` is used in `software/colorHarmonize/CMakeLists.txt`_

pmoulon:
Feel free to help on this.

pmoulon:
> Also fast is missing from target_link_libraries in software/VO
Also lemon insted of emon is used in software/colorHarmonize/CMakeLists.txt

The CI does not complain about that.
Did you use a special setup?

bartoszek:
> fast should not be a problem.

Yes, there was my mistake. I drop all caches and `openmvg` rebuilds fine, without `fast.patch`. But ...

_colorHarmonize_
> The CI does not complain about that.
Did you use a special setup?

I'm using system `coin-or-lemon` package (v1.3.1) which provides both static and shared lib. When building with `lemon(internal)` there is no error, with `lemon(external)` i got this error form `make`:
```
/usr/bin/ld: cannot find -llemon
collect2: error: ld returned 1 exit status
make[2]: *** [software/colorHarmonize/CMakeFiles/openMVG_main_ColHarmonize.dir/build.make:134: Linux-x86_64-RELEASE/openMVG_main_ColHarmonize] Error 1
make[1]: *** [CMakeFiles/Makefile2:3782: software/colorHarmonize/CMakeFiles/openMVG_main_ColHarmonize.dir/all] Error 2
```
[...]

PR #1270, opened 2018-03-29
FindFlann.cmake using PkgConfig.

bartoszek:
This should do it #1265 
1. Move `OpenMVG_USE_INTERNAL_FLANN` inside `FindFlann.cmake`.
2. Append original `PkgConfig` script from flann to `FindFlann.cmake`
3. Add conditionals, fix missing variable, test, test, test.


pmoulon:
I will discard this PR since using `-DFLANN_INCLUDE_DIR_HINTS=/usr/include/flann` in the cmake command line is doing the trick.

Issue #1285, opened 2018-04-11
question about Similarity3 class


sebastienchappuis:
Hello !

I have a question regarding the `Similarity3` class and the `ApplySimilarity` function.

Intuitively, I would assume that applying a similarity and its inverse on a scene would lead to an identity transformation...but it seems like it is not the case ?

If I write some code like this : 
``` 
const int nviews = 3;
  const int npoints = 10;
  const nViewDatasetConfigurator config;
  const NViewDataSet d = NRealisticCamerasRing(nviews, npoints, config);

  SfM_Data sfm_data = getInputScene(d, config, PINHOLE_CAMERA);

  std::cout << "poses before similarity : " << std::endl;
  for (const auto & pose : sfm_data.GetPoses())
  {
    std::cout << pose.second.center() << std::endl;
    std::cout << "=======" << std::endl;
  }

  const geometry::Similarity3 sim = createArbitrarySimilarity();

  ApplySimilarity(sim, sfm_data);
  ApplySimilarity(sim.inverse(), sfm_data);

  std::cout << "poses after applying similarity and inverse similarity : " << std::endl;
  for (const auto & pose : sfm_data.GetPoses())
  {
    std::cout << pose.second.center() << std::endl;
    std::cout << "=======" << std::endl;
  }
```

the results I get are : 

```
poses centers before similarity : 
-1.29904
       0
   -0.75
=======
  0
  0
1.5
=======
1.29904
      0
  -0.75
=======
poses centers after applying similarity and inverse similarity : 
 -1.26374
-0.264706
-0.608824
=======
0.0352941
-0.264706
  1.64118
=======
  1.33433
-0.264706
-0.608824
=======
```

which is not what I would expect...
I'm not sure I fully understand the way it works maybe I'm missing something obvious here ?

thanks for your help !

sebastienchappuis:
hey @Yannnnnnnnnnnn, well this explains that :) I could write some unit tests on this if you want ?

sebastienchappuis:
so I applied your fix but it still does not work.

```
poses centers before similarity : 
-1.29904
       0
   -0.75
=======
  0
  0
1.5
=======
1.29904
      0
  -0.75
=======
poses centers after applying similarity and inverse similarity : 
  -1.17052
 -0.935274
-0.0717664
=======
  1.27574
-0.339104
-0.712382
=======
-0.105225
  1.27438
 0.784149
=======
```

pmoulon:
[...]

PR #1288, opened 2018-04-11
fix similarity scale error

yanqswhu:
There is an error in pose.inverse() when scale exists.

screenshot from 2018-04-11 21-21-14

### DIFF
--- a/src/openMVG/geometry/Similarity3.cpp
+++ b/src/openMVG/geometry/Similarity3.cpp
@@ -37,7 +37,7 @@ Pose3 Similarity3::operator () ( const Pose3 & pose ) const
 
 Similarity3 Similarity3::inverse() const
 {
-  return Similarity3(Pose3(pose_.rotation(),-pose_.rotation()*pose_.center()*scale_), 1.0 / scale_);
+  return Similarity3(Pose3(pose_.rotation().transpose(),-pose_.rotation()*pose_.center()*scale_), 1.0 / scale_);
 }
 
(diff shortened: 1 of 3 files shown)


src/openMVG/geometry/Similarity3.cpp
```diff
@@ -37,7 +37,7 @@ Pose3 Similarity3::operator () ( const Pose3 & pose ) const
 
 Similarity3 Similarity3::inverse() const
 {
-  return Similarity3(pose_.inverse(), 1.0 / scale_);
+  return Similarity3(Pose3(pose_.rotation(),-pose_.rotation()*pose_.center()*scale_), 1.0 / scale_);
```
NikolausDemmel: rotation needs to be inverted

sebastienchappuis: true, with the rotation inverted it solves #1285

pmoulon: @Yannnnnnnnnnnn  @sebastienchappuis Do you want to make a PR with a unit test and the fix?

Yannnnnnnnnnnn: made a stupid error here.

Yannnnnnnnnnnn: @pmoulon I will try to add an unit test here and fix it.

lionlai1989:
It seems this PR can be closed.

PR #1369, opened 2018-08-09
Corrected a problem with updating intrinsics after BA in localization…

hcjghr:
Hi @pmoulon,

while looking at the localization module I noticed that the intrinsics in the RefinePose function do not get updated correctly (they are not updated with the new optimized values). 

I solved the problem with a simple _updateFromParams_ call as the _shared_intrinsics_ is just a clone of _intrinsics_. I'm not sure if that is the best solution but please have a look.

Cheers,
Klemen


pmoulon:
I will have a look. Normally the two actions should do the same thing. Thx for noticing that.

pmoulon:
Thx. Fixed by 835f1e585377a55d4143548fa9181438f5d23889

PR #1372, opened 2018-08-16
Sfm export

chubei-oppen:
Changed sfm_data.bin to sfm_data_all.json, so we can modify sfm root path in pipeline script, thus allow moving data between machines.


chubei-oppen:
Sorry, didn't mean to merge to this repo.

Issue #1373, opened 2018-08-16
Huber loss width


NikolausDemmel:
I was woundering why in the BA implementation the Huber-loss parameter is chosen as it is: https://github.com/openMVG/openMVG/blob/db92617fea420363abcbad1cb3773706f75012ad/src/openMVG/sfm/sfm_data_BA_ceres.cpp#L330

Passing `Square(4.0)` looks like the intent is to choose 4 pixel width. This also would make sense since elsewhere observations with error > 4.0 pixel get removed (https://github.com/openMVG/openMVG/blob/db92617fea420363abcbad1cb3773706f75012ad/src/openMVG/sfm/pipelines/sequential/sequential_SfM.cpp#L147). However AFAICT ceres expects not the squared width, so as is the used kernel-width is 16 pixels.

I guess 4 pixels is just a coarse heutristics, or how was this number determined? I guess one can come up with more sophisticated ways to determine this, e.g. also taking into account the point-detector scale (points detected at coarser scale get wider huber-kernel width). Maybe in practice this doesn't matter to much, or what is your experience?

pmoulon:
Ceres use squared cost function, so the Loss should be squared too in order to match (the sqrt is skipped by ceres for speed issue).

4.0 has been determined experimentally.
For sure you can find better scaling scheme that could use the feature scale but I'm not sure it would change the results in a dramatic way.

You can try to look to what the other framework are doing... and make your own idea.

NikolausDemmel:
> Ceres use squared cost function, so the Loss should be squared too in order to match (the sqrt is skipped by ceres for speed issue).

In general yes, but in this particular case the documentation explicitly mentions that the parameter to Huber loss is "not squared" (it interally caches the squared parameter als `b_`).

https://github.com/openMVG/openMVG/blob/master/src/third_party/ceres-solver/include/ceres/loss_function.h#L159-L164

I can provide a patch to remove the squares if you want.

> 4.0 has been determined experimentally.
For sure you can find better scaling scheme that could use the feature scale but I'm not sure it would change the results in a dramatic way.

Thanks for the info. I agree that it probably doesn't make too much difference after all.

> You can try to look to what the other framework are doing... and make your own idea.

Just out of curiosity. COLMAP for example seems to use Cauchy loss with a fixed kernel width, which defaults to `1`:

[...]

PR #1393, opened 2018-09-13
remove two non consistent squares

rjanvier:
I believe these thresholds were squared by mistake. While I agree we could maybe raise the angular threshold, the squared residual error is far too permissive.


NikolausDemmel:
Related: #1373

rjanvier:
Yes I agree it's related to squaring thresholds, but here it's maybe more obvious than in the case of loss width.

NikolausDemmel:
Have you checked if the respective functions maybe expect squared thresholds?

rjanvier:
I know what I'm doing here ;). And an angle is unlikely to be squared.

pmoulon:
In fact here, the square was made on purpose.
Since here, we have an initial reconstruction (the 3D seed), I want to give more chance for the first round of resection to add new camera... Perhaps this is not a good idea. We can try this experimentally and see what is the best option (Conservative, or permissive and then conservative)...

PR #1394, opened 2018-09-14
[sfm] Improve robust triangulation

rjanvier:
Improve triangulation by:
- Using cheirality as an hard requierement for inlier classification
- Removing an useless test on exhaustive triangulation
 => in this case if one obs fail its obvious the number of valid observation will be < of the min_required_inliers_
- Adding a function to triangulate the full track in order to (hopefully) improve the computation time and simplify the code at some place (at the price of not being very DRY).

### DIFF
--- a/src/openMVG/sfm/sfm_data_triangulation.cpp
+++ b/src/openMVG/sfm/sfm_data_triangulation.cpp
@@ -40,6 +40,47 @@ SfM_Data_Structure_Computation_Blind::SfM_Data_Structure_Computation_Blind
 {
 }
 
+/// Triangulate a given track
+bool track_full_triangulation
+(
+  const SfM_Data & sfm_data,
+  const Observations & obs,
+  Vec3 & X
+)
+{
+  if (obs.size() >= 2)
+  {
+    std::vector<Vec3> bearing;
+    std::vector<Mat34> poses;
+    bearing.reserve(obs.size());
+    poses.reserve(obs.size());
+    for (const auto& curr_obs : obs)
+    {
+      const View * view = sfm_data.views.at(curr_obs.first).get();
+      if (!sfm_data.IsPoseAndIntrinsicDefined(view))
+        continue;
+      const IntrinsicBase * cam = sfm_data.GetIntrinsics().at(view->id_intrinsic).get();
+      const Pose3 pose = sfm_data.GetPoseOrDie(view);
+      bearing.emplace_back((*cam)(cam->get_ud_pixel(curr_obs.second.x)));
+      poses.emplace_back(pose.asMatrix());
+    }
+    if (bearing.size() >= 2)
+    {
+      const Eigen::Map<const Mat3X> bearing_matrix(bearing[0].data(), 3, bearing.size());
+      Vec4 Xhomogeneous;
+      TriangulateNViewAlgebraic
+      (
+        bearing_matrix,
+        poses, // Ps are projective cameras.
+        &Xhomogeneous);
+      X = Xhomogeneous.hnormalized();
+      return true;
+    }
+  }
+  return false;
+}
+
+
 /// Triangulate a given track from a selection of observations
 bool track_sample_triangulation
 (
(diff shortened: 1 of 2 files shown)


rjanvier:
@pmoulon 

https://github.com/openMVG/openMVG/blob/76bca0bcbfff07bdcf11161ef294904f312d3c1a/src/openMVG/sfm/sfm_data_triangulation.cpp#L345-L346

Why are we converting the vector into a set here? Could the robust::UniformSample function yields some duplicate indices?

rjanvier:
Track_sample_triangulation take a set but I think it doesn’t really need one internally, so a vector should be ok here.

rjanvier:
I think this PR is complete now

src/openMVG/sfm/sfm_data_triangulation.cpp
```diff
@@ -40,12 +40,53 @@ SfM_Data_Structure_Computation_Blind::SfM_Data_Structure_Computation_Blind
 {
 }
 
+/// Triangulate a given track
+bool track_full_triangulation
```
pmoulon: `track_full_triangulation` and `track_sample_triangulation` should share code.
I think in an ideal world `track_full_triangulation` should call `track_sample_triangulation` and just create the samples [0, ..., n].
I would prefer minimal code to maintain rather than code duplicates with minor modification.
We can then measure in a second time the speed improvement and memory requirements.

Else it means that the design is bad that we need a ObservationsSampler structure that will use a reference to a whole Observations or create a sub-copy or reference with the right value on the flight if there is a sampling.


Perhaps would be to only have one function `track_obs_triangulation(sfm_data, obs, X)` that we could call also with: `track_obs_triangulation(sfm_data, ObservationSampler(obs, sampler_idx), X)`

src/openMVG/sfm/sfm_data_triangulation.cpp
```diff
@@ -40,12 +40,53 @@ SfM_Data_Structure_Computation_Blind::SfM_Data_Structure_Computation_Blind
...
+}
+
+
 /// Triangulate a given track from a selection of observations
 bool track_sample_triangulation
 (
   const SfM_Data & sfm_data,
   const Observations & obs,
-  const std::set<IndexT> & samples,
+  const std::vector<IndexT> & samples,
```
pmoulon: We should document that we expect to do not duplicate in the samples array.
`const std::vector<IndexT> & samples, // Unique ids`

src/openMVG/sfm/sfm_data_triangulation.cpp
```diff
@@ -335,13 +365,13 @@ const
...
       const Pose3 pose = sfm_data.GetPoseOrDie(view);
-      bChierality &= CheiralityTest((*cam)(itObs->second.x), pose, X);;
+      bCheirality &= CheiralityTest((*cam)(itObs->second.x), pose, X);
       const Vec2 residual = cam->residual(pose(X), itObs->second.x);
       bReprojection_error &= residual.squaredNorm() < dSquared_pixel_threshold;
-      validity_test_count += (bChierality && bReprojection_error) ? 1 : 0;
+      validity_test_count += (bCheirality && bReprojection_error) ? 1 : 0;
     }
 
-    if (!bChierality || !bReprojection_error ||
```
pmoulon: I have made the typo `chierality` instead of `cheirality` in 24 places across the project (my bad)

src/openMVG/sfm/sfm_data_triangulation.cpp
rjanvier: I agree, I could try to explore the ObservationSampler idea if you want, it's elegant.

src/openMVG/sfm/sfm_data_triangulation.cpp
rjanvier: I never payed attention before, "Chierality" sounds good in french.

src/openMVG/sfm/sfm_data_triangulation.cpp
pmoulon: I'm doing it right now. I will certainly able to push to develop this week end.

src/openMVG/sfm/sfm_data_triangulation.cpp
pmoulon: :fr: ;-)

pmoulon:
@rjanvier I used your branch as a starting point and add a new design to test hypothesis on a track using a std::function. It allows to customize the check on the track. This way we can check the track for cheirality or cheirality and residual error.
See here:
https://github.com/openMVG/openMVG/commit/6ac3d39e82226f82d1fa433fb60d6f047731106e#diff-cf336d3a39474c8f479fc238f4f4eceeR95

Let me know if you like this design and if everything looks good for you.

I also add dedicated test for testing the sfm_data_triangulation engines. The two engines were already tested thanks to the SfMPipeline but it is always good to add some unit test.

PR #1399, opened 2018-09-20
[SfMv2] reset track ratio each time we augment the pose set

rjanvier:
the track_inlier_ratio is  a good safe guard to avoid to estimate poses of yet weakly supported views.

Nevertheless, actually, once we reach a part of the view graph with a weakly supported views the track_inlier_ratio is downgraded (from 0.2 to 0.0) and it is never upgraded again until the end of the reconstruction. Weakly supported part of the viewgraph could be transition between two "communities" and thus I think we should reset the ratio each time we augment the pose set (I do think it was the intended behavior) to avoid to register some view too early that could create a "very unfortunate ordering". On my experiments, reseting the track_inlier_ratio leads to **really** far more stable reconstructions (less bad registrations), at the expense of course of a bigger number of iterations (but maybe we could fine tune the ratios. something like this works well on my sets { 0.15, 0.05, 0.02, 0.0 }).

### DIFF
--- a/src/openMVG/sfm/pipelines/sequential/sequential_SfM2.cpp
+++ b/src/openMVG/sfm/pipelines/sequential/sequential_SfM2.cpp
@@ -148,8 +148,8 @@ bool SequentialSfMReconstructionEngine2::Process() {
 
   // Incrementally estimate the pose of the cameras based on a confidence score.
   // The confidence score is based on the track_inlier_ratio.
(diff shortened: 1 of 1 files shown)


pmoulon:
Thanks. Merged by https://github.com/openMVG/openMVG/commit/0f679d9f06b354871ff6ea7bc012eeeeb68c542d

PR #1414, opened 2018-10-11
[cameras] simplify AngleBetweenRay

rjanvier:
Since `ray1` and `ray2` are already normalized` ray1.norm() `== `ray2.norm()` = 1


pmoulon:
LGTM, I will merge during the week end ;-) Thx!

PR #1476, opened 2019-02-11
Implementation of Llyod's kmeans clustering algorithm

pmoulon:


pmoulon:
Pushed in https://github.com/openMVG/openMVG/tree/rperrot-kmeans for further investigation

PR #1486, opened 2019-03-07
Split SFM Compute matches into functional parts

rperrot:
Hi, 

this PR aims at splitting the openMVG_main_ComputeMatches executable into three executables: 
- openMVG_main_PairGenerator : given an sfm_data, it computes the set of image pairs to be matched. It comes from the original openMVG_main_ComputeMatches executable where two modes are available: exhaustif pair (ie: all possible combination of image pairs) and contiguous (ex: video mode)
- openMVG_main_ComputeMatches : it now does only matching.
- openMVG_main_GeometricFilter : it does geometric only filtering, it accepts as input a matches file. It comes from the original openMVG_main_ComputeMatches file. 

The standard pipelines have been updated to take into account this new division. 

The new executables may provide more flexibility for hybrid pipelines because it allows multiple pass.

Hope it helps !

### DIFF
--- a/src/openMVG/matching/indMatch.hpp
+++ b/src/openMVG/matching/indMatch.hpp
@@ -102,6 +102,26 @@ inline Pair_Set getPairs(const PairWiseMatches & matches)
   return pairs;
 }
 
+/**
+ * @brief Get the subset of the matches that corresponds to the given pairs 
+ * 
+ * @param matches     Initial matches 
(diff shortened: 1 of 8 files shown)


pmoulon:
Thx you Romuald 🎉 , I'm adding here the new visual flow:
screen shot 2019-03-07 at 10 29 11 pm

pmoulon:
Already merged

PR #1491, opened 2019-03-20
Find external Eigen

tkonolige:
OpenMVG fails to find an external eigen because `find_package(eigen3)` should be `find_package(Eigen3)`.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -242,8 +242,8 @@ endif()
 # - external if EIGEN_INCLUDE_DIR_HINTS is defined
 # - internal if Eigen not found
 # ==============================================================================
-find_package(eigen3 QUIET)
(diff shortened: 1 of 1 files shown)


pmoulon:
Thx, building again to check the VS studio CI build.

tkonolige:
Is the VS studio build error related to this change? It looks like it makes it past the make phase.

pmoulon:
Most of them are working ok.
I think the free plan is not really stable and sometimes some build are failing for a strange reason.
```MSBUILD : error MSB4017: The build stopped unexpectedly because of an unexpected logger failure.
System.Net.WebException: The underlying connection was closed: A connection that was expected to be kept alive was closed by the server.```

I will merge your change to develop, and then when the next version will be released it will be pushed in master ;-)

PR #1500, opened 2019-04-05
[Multiview] Add lambda twist p3p solver

rjanvier:
cc Mikael Persson (@midjji)

This PR contains the code the of the P3P method exposed in Persson M., Nordberg K., Lambda Twist: An Accurate Fast Robust Perspective Three Point (P3P) Solver, ECCV 2018.

- I had just adapted the code present in the original repository (https://github.com/midjji/lambdatwist-p3p) to Eigen/openMVG. I kept some comments and maybe some duplicated methods w.r.t what we already have in openMVG. I think further refinements, if needed, are out of my league and maybe @midjji and @pmoulon you could now further discuss about how to best integrate lambda twist in openMVG.

- According to discussion we had with Mikael, the code is now MPL-2 licensed (instead of the original GPL-3.0 scheme) and the method is named in honor of Klas Nordberg.
  
The implementation is tested and functional. I made some extensive real life tests, using it for my main solver in several dataset (one was about 7790 HR images of Château de Chambord ;)). It's very stable and plays nicely with AC-RANSAC. I think it should be default p3p solver in openMVG.

Thank you so much to Klas Nordberg and Mikael Persson for making this code public under a permissive license.

### DIFF
--- a/src/openMVG/multiview/solver_resection_p3p_nordberg.cpp
+++ b/src/openMVG/multiview/solver_resection_p3p_nordberg.cpp
@@ -3,7 +3,9 @@
 // Copyright (c) 2018 Michael Persson
 // Adapted to openMVG by Romain Janvier
 
-// This source code is subject to the GPL 3.0 licence
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
 #include "openMVG/multiview/solver_resection_p3p_persson.hpp"
 #include "openMVG/multiview/projection.hpp"
 
--- a/src/openMVG/multiview/solver_resection_p3p_nordberg.cpp
+++ b/src/openMVG/multiview/solver_resection_p3p_nordberg.cpp
@@ -271,7 +271,6 @@ void P3PSolver_Nordberg::Solve(
       models->push_back(P);
     }
   }
-  Mat solutions = Mat(3, 4 * 4);
 };
 
 double P3PSolver_Nordberg::Error
--- a/src/openMVG/multiview/solver_resection_p3p_nordberg.cpp
+++ b/src/openMVG/multiview/solver_resection_p3p_nordberg.cpp
@@ -520,7 +520,6 @@ void P3PSolver_Nordberg::Solve(
   assert(3 == bearing_vectors.rows());
   assert(3 == X.rows());
   assert(bearing_vectors.cols() == X.cols());
-  Mat34 P;
   std::vector<std::tuple<Mat3, Vec3>> rotation_translation_solutions;
   if (computePosesNordberg(bearing_vectors, X, rotation_translation_solutions))
   {
--- a/src/openMVG/multiview/solver_resection_p3p_nordberg.cpp
+++ b/src/openMVG/multiview/solver_resection_p3p_nordberg.cpp
@@ -96,20 +96,20 @@ static void gauss_newton_refineL(Vec3 &L,
 
 static inline bool root2real(const double & b, const double & c, double & r1, double & r2)
 {
-    double v = b * b -4.0 * c;
-    if (v < 0.0) {
-        r1 = r2 = 0.5 * b;
-        return false;
-    }
-    double y = std::sqrt(v);
-    if (b < 0.0) {
-        r1 = 0.5 * (-b + y);
-        r2 = 0.5 * (-b - y);
-    } else {
-        r1 = 2.0 * c / (-b + y);
-        r2 = 2.0 * c / (-b - y);
-    }
-    return true;
+  double v = b * b -4.0 * c;
+  if (v < 0.0) {
+      r1 = r2 = 0.5 * b;
+      return false;
+  }
+  double y = std::sqrt(v);
+  if (b < 0.0) {
+      r1 = 0.5 * (-b + y);
+      r2 = 0.5 * (-b - y);
+  } else {
+      r1 = 2.0 * c / (-b + y);
+      r2 = 2.0 * c / (-b - y);
+  }
+  return true;
 };
 
 /**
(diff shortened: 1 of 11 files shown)


pmoulon:
Thank you guys. I will check why we have a compilation error for CLANG and VSTUDIO ;-) and make a code review.

src/openMVG/multiview/solver_resection_p3p_nordberg.cpp
```diff
@@ -0,0 +1,289 @@
...
+    for (const auto & rot_trans_it : rotation_translation_solutions) {
+      Mat34 P;
+      P_From_KRt(Mat3::Identity(),           // intrinsics
+                  std::get<0>(rot_trans_it), // rotation
+                  std::get<1>(rot_trans_it), // translation
+                  &P);
+      models->push_back(P);
+    }
+  }
+  Mat solutions = Mat(3, 4 * 4);
```
pmoulon: delete this line.

src/openMVG/multiview/solver_resection_p3p_nordberg.hpp
```diff
@@ -0,0 +1,248 @@
...
+    if(v < 0.0){
+        r1 = r2 = 0.5 * b;
+        return false;
+    }
+    double y = std::sqrt(v);
+    if(b < 0.0){
+        r1 = 0.5 * (-b +y);
+        r2 = 0.5 * (-b -y);
+    }else{
+        r1= 2.0 * c / (-b + y);
```
pmoulon: minor style issue `r1 = `

src/openMVG/multiview/solver_resection_p3p_nordberg.hpp
```diff
@@ -0,0 +1,248 @@
...
+        r2= 2.0 * c / (-b - y);
```
pmoulon: same for r2

src/openMVG/multiview/solver_resection_p3p_nordberg.hpp
```diff
@@ -0,0 +1,248 @@
...
+    }
+    return true;
+};
+
+static double cubick(const double & b, const double & c, const double & d){
```
pmoulon: We already have a cubic solver, we could leverage the existing if we can https://github.com/openMVG/openMVG/blob/835f1e585377a55d4143548fa9181438f5d23889/src/openMVG/numeric/poly.h#L32

src/openMVG/multiview/solver_resection_p3p_nordberg.hpp
```diff
@@ -0,0 +1,248 @@
...
+};
+
+//-- Usable solver for robust estimation framework
+using PoseResectionKernel_P3P_Persson =
+    two_view::kernel::Kernel<
+        P3PSolver_Nordberg, // Model estimator
+        P3PSolver_Nordberg, // Error metric
+        Mat34>;
+
+static void gauss_newton_refineL(Vec3 &L,
```
pmoulon: I would move all the method below to the CPP.

src/openMVG/multiview/solver_resection_p3p_nordberg.cpp
rjanvier: ouch an useless alloc! nice catch! thanks!

src/openMVG/multiview/solver_resection_p3p_nordberg.hpp
rjanvier: thanks! I saw a couple of missing spaces in others lines.

src/openMVG/multiview/solver_resection_p3p_nordberg.hpp
rjanvier: I was not sure about that. Is there any risk to loose (or gain) accuracy or are the two solvers equivalents? (On my side, I'm not very concerned about speed as it's already very fast)

src/openMVG/multiview/solver_resection_p3p_nordberg.hpp
rjanvier: and also the root2real one + eigwithknown0 one?

src/openMVG/multiview/solver_resection_p3p_nordberg.hpp
pmoulon: It seems that both solver as a bit different after looking in details (1 vs 3 roots).
My recommendation is moving those couple of function to the cpp would be good to avoid having other interface using the functions.

src/openMVG/multiview/solver_resection_p3p_nordberg.hpp
pmoulon: yep. I would love having only the solver in the HPP. Since all the other function are for internal use

src/openMVG/multiview/solver_resection_p3p_nordberg.hpp
rjanvier: yes I reread the paper after 6 month, it seems to one of the advantage of the method over others p3p solvers.

midjji:
Hi,

It is crucial that the 1 root cubic solver is used. This impacts both speed
and accuracy substantially compared to using any standard 3 root cubic
solver.

The difference is caused by the numeric instability of the common higher
multiplicity roots etc.

The same thing is true for the specialized eigenvalue decomp. It cannot be
replaced by a generic component without both speed and accuracy loss.

The second order root solver isn't important.

With regards to unnecessary allocation I did mention that std vector should
not be used, and it isn't in the original. Prealloced STD vectors shouldn't
be used either because they still break alot of compiler optimisations.
Cheers
//Mikael

src/openMVG/multiview/solver_resection_p3p_nordberg.cpp
```diff
@@ -0,0 +1,542 @@
...
+* @param[in] X  3x3 matrix with corresponding 3D world points (each column is a point)
+* @param[out] rotation_translation_solutions vector that will contain the solutions (up to 4 solutions)
+*
+* @return true if at least one solution is found, false if no solution was found
+*
+*/
+bool computePosesNordberg(
+    const Mat &bearing_vectors,
+    const Mat &X,
+    std::vector<std::tuple<Mat3, Vec3>> &rotation_translation_solutions)
```
rjanvier: @midjji you mean this std::vector?

src/openMVG/multiview/solver_resection_p3p_nordberg.cpp
```diff
@@ -0,0 +1,542 @@
...
+  // get sorted eigenvalues and eigenvectors given that one should be zero...
+  Mat3 V;
+  Vec3 L;
+
+  eigwithknown0(A, V, L);
+
+  double v = std::sqrt(std::max(0.0, -L(1) / L(0)));
+
+  int valid = 0;
+  std::array<Vec3, 4> Ls;
```
rjanvier: @midjji or maybe the use of the std::array here?

midjji:
The first own, the vector.

rjanvier:
> The first own, the vector.
@midjji 
Ok, I kept the interface we have for Ke's solver. I will stick to your implementation then.

midjji:
Took a look at the 3 root cubic solver linked,

While cos formulation isn't the best numerically in general.
I think you also have specific numerical issues for q~0 in the std::arcos,
this is especially bad since a cubic solver should never throw a NAN. the
main problem is that the accuracy of p^3 isn't guaranteed to be caught by
the numeric check on p and the threshold probably needs to be increased.
The input must also have been explicitly normalized, which is unwise, it
requires the user to properly account for the special cases which is ...
unlikely. Performing that division is also not required, and doing it in
advance actually reduces numerical accuracy compared to keeping it in the
formulas.

Throw a few billion random polynomials at it and id wager you'll get a few
nans and quite a few bad solutions near roots of higher multiplicity. Might
want to explicitly test bad cases like (x-0.99)(x-1 +-1e-8) too.

There are specializations many cases you'll want to add for robustness, but
I only know of a few,
but most importantly, the roots would gain several bits of accuracy from
refinement.

Not sure if any of this is important, but at least it might be good to be
aware.

pmoulon:
@midjji Thank you for your feedback.

I agree that each solver should use the best numerical precision and sometimes it forces to uses a specific root solver that could be optimized for speed and accuracy. We will keep your specific root solver. I was trying to see if we could factorize some code. We are glad to have your feedback.

Regarding vector vs array. In order to keep some genericity due to the fact that some solver can return multiple solutions, we need vector at the end. So std::array could be used inside the solver and then only a single solution will be returned in your case.

rjanvier:
@pmoulon it returns up to 4 solution like any other p3p solver.

so the question is : should we keep the `std::vector` (for sake of consistency with the other p3p interfaces in openMVG) or use arrays (to be truly fair with the original code and in order to maximize speed) in `computePosesNordberg`? 
Even if i think the speed is not an issue for us, since openMVG does not target (primarly) real time applications, I think it's not a big dealhere  for openMVG if we stick with the original code, since the call of `computePosesNordberg` is wrapped inside another function anyway (`P3PSolver_Nordberg::solve`) and it hides the fact we use preallocated arrays vs. dynamic vectors.

RamadanAhmed:
I think you should add #include<array> in solver_resection_p3p_nordberg.cpp to compile in visual studio

rjanvier:
> 
> 
> I think you should add #include in solver_resection_p3p_nordberg.cpp to compile in visual studio

Thanks @RamadanAhmed have you any idea of the specific missing include?

midjji:
I agree that the Interface compatability forces the use of the vector, it's
pretty DN... convenient too. The performance increase will be minor as long
as the vector is prealloced to 4. Romain is right that it will return
multiple valid solutions, generally fewer total but more distinct ones than
the other solvers too.

Is the use of rotation matrixes important btw? I only used that for fair
speed comparisons with ke and kneip, the unit quaternion version is
slightly  better.

RamadanAhmed:
> 
> 
> > I think you should add #include in solver_resection_p3p_nordberg.cpp to compile in visual studio
> 
> Thanks @RamadanAhmed have you any idea of the specific missing include?

#include <array> sorry I forget to mention :)

rjanvier:
Ok if I understand well we keep the vector in the `computePosesNordberg``? 
For the rotation we use a the matrix parametrisation in openMVG so it's that we want!

pmoulon:
@rjanvier I agree with your comment https://github.com/openMVG/openMVG/pull/1500#issuecomment-483581519
- Sorry if I was not clear: Using the array in the computePosesNordberg is totally fine. We have to keep the vector<> only in the generic interface that will be plugged into the kernel for robust estimation.

Congrats team we are very close to having another minimal solver in Develop to be tested! 🎉

RamadanAhmed:
I have also tested it and it works much better than P3P_Ke

pmoulon:
Thank you @RamadanAhmed for testing the WIP feature. Appreciated!

- Once I will merge to develop, once it is tested, I will make it default.

We just need to test its behavior for pinhole and 360 images.

@rjanvier Seems like a minor concern is still there to the variable fx here https://app.codacy.com/app/pmoulon/openMVG/pullRequest?prid=3389059 (I can fix it)

rjanvier:
@pmoulon I think allocating fx/fpx into the loop could impact the perfomance... AFAIK it really depends on the compiler so I haven't fixed it on purpose.

midjji:
It could, but in practice dont worry about it for recent compilers in this
specific case.

As you might have noticed, I chased every last cycle and every last bit of
accuracy.
And I wanted to do so in a way which is cross architecture friendly, so no
explicit asm.
The latter means a bunch of strange choices in places and alot of godbolt.

midjji:
360 images? hmm might not work for that, whats the projection model?
Let me know if it does or not, if it doesnt that be great, since then I can
fix that.

rjanvier:
@midjji it's an equirectangular projection model.  I did a test yesterday with 360 images, lambda twist works at least as well as Ke's solver (regarding behavior). We use bearing vectors as input (not 2 coordinates in sensor's frame) and it seems you use a bearing vector formulation too in the solver, so it's general enough but maybe I'm wrong and I'm missing some important points here, please let me know. thanks a lot!

rjanvier:
Ok my bad, it seems that the spherical model violate the assumption that all points must have the same signed distance w.r.t to the camera center. Am I right? 
It seems that it worked well on my spherical dataset but it's certainly due to the fact I have always 3 "good" 3D points lying in one hemisphere... I made a simple synthetic test with 3 points well distributed on the sphere and it shows that it's not working. You said you have a possible "fix" for this case? Else it's not a big deal, I think we can easily restrict the use of lambda twist to the pinhole model and keep Ke's solver for spherical cameras.

rjanvier:
Ok my bad, it seems that the spherical model violate the assumption that all points must have the same signed distance w.r.t to the camera center. Am I right? 
It seems that it worked well on my spherical dataset but it's certainly due to the fact I have always 3 "good" points lying in one hemisphere... I made a simple synthetic test with 3 points well distributed on the sphere and it shows that it's not working. You said you have a possible "fix"? Else it's not a big deal, I think we can easily restrict the use of lambda twist to the pinhole model and keep Ke's solver for spherical cameras.

rjanvier:
forgot my last comment, my synthetic test was wrong.

src/openMVG/multiview/solver_resection_p3p_nordberg.cpp
```diff
@@ -0,0 +1,546 @@
...
+  Vec3 P1 = X.col(0);
+  Vec3 P2 = X.col(1);
+  Vec3 P3 = X.col(2);
+
+  // Extraction of feature vectors
+  Vec3 f1 = bearing_vectors.col(0);
+  Vec3 f2 = bearing_vectors.col(1);
+  Vec3 f3 = bearing_vectors.col(2);
+
+  f1.normalize();
```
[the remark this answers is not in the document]
pmoulon: Yes they are.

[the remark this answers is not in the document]
rjanvier: thanks

midjji:
I'm unclear on if the 360 case worked or not?

src/openMVG/multiview/solver_resection_p3p_nordberg.cpp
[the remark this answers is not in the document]
rjanvier: @pmoulon in fact they are not, at least in pinhole camera model the operator() return return `Kinv_ * points.colwise().homogeneous();` so the point is expressed in homogenous coordinates and it's not an unit vector (will be `v(somescalar, somescalar,1)`).

midjji:
I left in explicit L2 normalization, because its one of those things which
while required, might accidentally work in many cases, causing annoyring to
debug errors.

rjanvier:
@midjji 360 camera works on my experiments (both real cases and synthetics). I'm not 100% sure though because I do not have a strong mathematical background (I came from an "humanities" background) and I'm not able to extract by myself the "proof" from your code/paper. 
Moreover in the paper one of the prerequisite of your methods is "since the parametersλkis the signed distance from the camera center for each 3D point,λ1, λ2,andλ3 are required to be positive and real. This is a geometric feasibility condition on the solutions, which implies that all three 3D points are “in front of” the camera." and it seems to contradict the way spherical projection work and my empirical observations ;).
Regarding the L2 normalization, your are right, I will revert the commit where I removed it because I assumed we use unit vectors in openMVG ("bearing vectors") but I just saw that in the case of pinhole camera we still use straight homogeneous coordinates.

pmoulon:
Merged by https://github.com/openMVG/openMVG/commit/b09174aab61e0f3ee726df197c47eade566ad52a and https://github.com/openMVG/openMVG/commit/0a4faee3db88dde8b8ebdf0aff7a5c4ddabcd890

Thank you @midjji, @rjanvier 

@RamadanAhmed This feature is ready to test, feel free to try it in the develop branch and provide any feedback.

Issue #1255, opened 2018-03-15
Strange behavior of IncrementalSfM (Resection part) when using GPS priors (-P)


hcjghr:
Hi,

From the beginning I apologize for a bit longer post:) 

In recent weeks I have been noticing that the IncrementalSfM has been acting strange in the resection step when motion prior switch (-P) is set. After the first BA with motion priors resection thresholds increase significantly causing problems in subsequent steps.

To illustrate the problem I have run the same IncrementalSfM command (with and without -P) on the same feature and match files.

In both cases the process starts (nearly) the same, with same initial pair (forced by -a and -b) and then followed by a series of robust resections for new images. The robust resection threshold is varies between the images but is reasonable (around 3-4) while final RMSE after the single pose BA below 1px.

The behavior is the similar in both cases until the first complete BA (after 19 images for example with P and after 25 images without P)

**- with P**

Bundle Adjustment statistics (approximated RMSE):
 #views: 174
 #poses: 19
 #intrinsics: 1
 #tracks: 50840
 #residuals: 438129
 Initial RMSE: 0.305092
 Final RMSE: 0.188702
 Time (s): 17.3881

Usable motion priors: 1
Pose prior statistics (user units):
 - Starting median fitting error: 0.0649667
 - Final fitting error:
	 min: 0.0315291
	 mean: 0.096793
	 median: 0.0699325
	 max: 0.273009

**- without P**

Bundle Adjustment statistics (approximated RMSE):
 #views: 174
 #poses: 26
 #intrinsics: 1
 #tracks: 61845
 #residuals: 596120
 Initial RMSE: 0.218907
 Final RMSE: 0.214843
 Time (s): 6.6084

I'm assuming that the difference between would be due to the randomness in the AC-RANSAC step (does that make sense?). Also, as the motion priors are not exact the error in the alignment is in the expected range.

**The big difference occurs after the first BA with motion priors as the final thresholds of the robust resection are differ for an order of magnitude.**

**- with P**

-- Robust Resection of view: 39
  nfa=27.4825 inliers=7/8445 precisionNormalized=0.0269952 precision=229.827 (iter=0 ,sample=6880,1144,7649,7052,1075,8182,)
  nfa=26.3097 inliers=7/8445 precisionNormalized=0.00181329 precision=59.5649 (iter=2 ,sample=2351,1591,4619,8384,8086,8415,)
[...]

PR #1513, opened 2019-04-24
[camera] use normalized bearing vector instead of homogenous coords

rjanvier:
Normalized bearing are needed by p3p solvers (Ke, and Kneip), currently we feed them with homogenous coordinates, it provokes drift in AC-RANSAC thresholds (see #1255).

it fixes at least #1255


pmoulon:
Thx you @rjanvier

PR #1518, opened 2019-04-29
[Software] Fix main_evalQuality

rjanvier:
Ground truth was compared against itself...


pmoulon:
My bad, introduced this by debugging the code last time.

PR #1520, opened 2019-05-06
[SfM/triangulation] the whole track was checked against predicates (not only the hypothesis)

rjanvier:
In the `SfM_Data_Structure_Computation_Robust` class: 
Since the inception of the function `track_check_predicate`, the full track was passed to the function (instead of checking *only* the hypothesis). As a result some valid hypothesis are discarded and the robust triangulation create less "valid" tracks than it should.
My bad for not having spotted this before the merge :-1:

### DIFF
--- a/src/openMVG/sfm/sfm_data_triangulation.cpp
+++ b/src/openMVG/sfm/sfm_data_triangulation.cpp
@@ -359,13 +359,15 @@ const
     std::vector<uint32_t> samples;
     robust::UniformSample(min_sample_index_, obs.size(), random_generator, &samples);
 
-    // Hypothesis generation
     Vec3 X;
-    if (!track_triangulation(sfm_data, ObservationsSampler(obs, samples), X))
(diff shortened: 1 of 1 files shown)


pmoulon:
link https://github.com/openMVG/openMVG/pull/1394/files#diff-cf336d3a39474c8f479fc238f4f4eceeR297 for history.
We could add a unit test to catch this. I will look to it tomorrow.

pmoulon:
Thx @rjanvier 

Merged by https://github.com/openMVG/openMVG/commit/bf6e2be92dc03f922775503db445130728deffc9.
Added a unit test to catch if we are doing something wrong in the future :smile: https://github.com/openMVG/openMVG/commit/eb181627fdf91abe610b8be06624478fd262ee20

I also increased the number of Ransac iteration for the robust triangulations in order to avoid to get stuck on sampling an outlier too often. Let me know what you think...

PR #1534, opened 2019-05-28
[software] fix main_ConvertSfM_DataFormat.cpp

rjanvier:
-C flag was ineffective


pmoulon:
Thank you @rjanvier 
`Errare humanum est`

PR #1569, opened 2019-07-31
[multiview/P3P] correct the GN solver used in Nordberg P3P

rjanvier:
May improve speed
see https://github.com/midjji/lambdatwist-p3p/commit/c1bb3ad398119d241ff6db5fa289446305102762
and https://github.com/midjji/lambdatwist-p3p/issues/2


pmoulon:
Thx you @rjanvier

PR #1570, opened 2019-08-01
Defined but Unused threshold parameter maxContrast, adding usage of threshold parameter to code,tiny fix

Mirmix:
While working on the KVLD filtering I find out that one of the parameter named maxContrast was defined but not used. I made a tiny change  in the code to put the threshold check to its right place.
Since the lines with high contrast might be unreliable, in particular along the image edges it would be better to have this threshold parameter and use it.

### DIFF
--- a/src/openMVG/matching/kvld/kvld.h
+++ b/src/openMVG/matching/kvld/kvld.h
@@ -103,7 +103,7 @@ public:
     diff[ 0 ] = 0;
     diff[ 1 ] = 0;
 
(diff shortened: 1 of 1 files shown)


pmoulon:
Thank you @Mirmix.
Appreciated that you saw that and share the fix!

pmoulon:
Just for your information, I prefer people to propose to fix towards the `develop` branch.
Master is used only to show the latest state of the last release. see [here](https://github.com/openMVG/openMVG/blob/master/CONTRIBUTING.md)

I will fix the destination branch on Github.

PR #1586, opened 2019-08-15
Small typo in documentation

Mirmix:
There was a tiny typo: ou -> or
I also added the world non-coplanar since correspondences are likely and expected to be not on the same plane, otherwise it creates degeneracy which makes it impossible to correctly calculate F matrix.
Minimum number correspondences required is 7. But since it was already 8 ou 7 , i just changed it to 8 or 7.


pmoulon:
Thank you. I'm french and so I made the typo because `or` is `ou`...

PR #1599, opened 2019-09-12
[Mutliview/P3P] fix #1597 thx @mpizenberg & @midjji

### LINKED ISSUE #1597: Potential issue in lambda twist p3p inclusion to OpenMVG
@rjanvier 

Hi, I'm currently porting the lambda twist p3p algorithm to Rust and so first, I'm basically rewriting `solver_resection_p3p_nordberg.cpp` into a compilable Rust version. While reading the `gauss_newton_refineL` function I think I spotted a mistake that makes this function potentially wrong.

At [lines 80-82](https://github.com/openMVG/openMVG/blob/master/src/openMVG/multiview/solver_resection_p3p_nordberg.cpp#L80-L82), `l1, l2, l3` are recomputed using the same unchanged `L` vector. This means that the new residual is always the same that the old one and the condition at [line 86](https://github.com/openMVG/openMVG/blob/master/src/openMVG/multiview/solver_resection_p3p_nordberg.cpp#L86) is therefore always false, leading to `L` always being updated by the new lambda `L1`, even in cases where the residual increases.

Let me know if I'm mistaken, I'm not very familiar with the algorithm but this stroke me as something weird.

rjanvier:


pmoulon:
Thx you @rjanvier I'm gonna wait that you add the remaining quadrant fix to this branch so we will have one commit for all the update.

pmoulon:
Merged. Thx

Issue #1618, opened 2019-10-24
indMatch.hpp file suggestion


GabrielMachine:
"Hello!
I  would like to report a suggestion in indMatch.hpp file. I was studying the OpenMVG for my final course project and I found in this file three friend bool operators. As far as I know these friend tokens are used when you have encapsulated  data in the class, but the data that is pretented to get these functions are public. I've also tested if removing this friend operator in the class would affect the code itself, however it does not affect and OpenMVG builds and runs as usual. So, I propose that  you remove the friend token. If desired, I can provide a pull request."

GabrielMachine:
I forgot to mention witch lines should be altered. the lines are line 29, line 33 and line38

pmoulon:
Thank you @GabrielMachine for your feedback.
You are right, the `friend` keyword is not needed since the object `IndMatch` is a `struct` and so I can have direct access to the struct member by using functions.
https://github.com/openMVG/openMVG/blob/master/src/openMVG/matching/indMatch.hpp#L29

Feel free to do a PR starting from develop branch and propose the PR to OpenMVG::develop branch. Thank you.

Perhaps I made the same mistake in some other structure. Feel free to make a search and see if we can fix some other operator ;-)

Feel free to explain what did you achieve with OpenMVG for your final course project or share any screenshot of nice reconstruction you made here ;-)

GabrielMachine:
I'm having an issue to send the pull request for the openMVG origin. It gives me the message "remote: Permission to openMVG/openMVG.git denied to GabrielMachine." . Could you provide me an way to solve this issue?

pmoulon:
You need to start from a FORK see https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/creating-a-pull-request

GabrielMachine:
I know this is doesn't seem urgent, but I am proposing what I feel could be the correct way for improving the API: I replaced the position of the operators to non-member  inline functions external to the struct and built it again and it worked.

See my new PR. I don't mean to bother but if you feel this should be done, I will do the same for all files and keep updating the PR until you feel it is OK.

Thank you.

PR #1619, opened 2019-10-24
closed(#1618)

GabrielMachine:


GabrielMachine:
I forgot to mention that this pull request closes the following issue:https://github.com/openMVG/openMVG/issues/1618

pmoulon:
Actually you see that the CI (Continuous Integration) is failing, so build is not working. See above `The Travis CI build failed`.
Did you try to build and run unit tests on your side before pushing a PR?

pmoulon:
If you want to remove FRIEND, you have to use member function and so you have to change some code:

```
  bool operator==(const IndMatch& m) const {
    return (this->i_ == m.i_ && this->j_ == m.j_);
  }

  bool operator!=(const IndMatch& m) const {
    return !(*this == m);
  }

  // Lexicographical ordering of matches. Used to remove duplicates
  bool operator<(const IndMatch& m) const {
    return (this->i_ < m.i_ || (this->i_ == m.i_ && this->j_ < m.j_));
  }

GabrielMachine:
Ok, I'll make these changes in file and test everything that have to be tested.

pmoulon:
If you are gonna fix this because we decide that friend function are not appropriate to have we should consider to fix all of them. See this https://github.com/openMVG/openMVG/search?utf8=%E2%9C%93&q=%22friend+bool%22&type= We have friend operator in various classes.  Personally I don't see this as an issue since it allow to use some of those operator as function operators.

But if we are gonna consider fixing it it here I want a uniform decision across the project.

@rperrot any comment on friend operator or member function?

GabrielMachine:
closing to try another solution

rperrot:
Hi, 

regarding friend keyword, it has never been an issue for operators (Core guidelines agrees with that). 

There are some issues with friend but it's not the case here:
- Long friend chip (accès to friend of a friend of a friend ...)
- friendchip that will break encapsulation too much (though it's not true because friend is only allowed if your class accept it) ;

The only thing that we may be prudent is :
- Automatic conversion may break behavior, with member function one of the operator can't be implicitly converted (on the first operand), with friend function it can happened. 

For operator== it we don't use class inheritance I think it's ok. But if inheritance and genericity occurs, we may prefer a function like "bool equal()" to make sure we don't do anything weird.

pmoulon:
@GabrielMachine Let's let the code as it is. I don't think we have a serious issue here.
Thank you for your feedback. Appreciated. It is always good to have a second look on the designed API and check we made the things in the right way.

Feel free to discuss with us if you want to contribute and help the project by writing here openmvg-team[AT]googlegroups.com

PR #1627, opened 2019-11-02
[continuous integration] added support for gcc 8

GabrielMachine:
Needed for any Ubuntu that have gcc 8 compiler as native (e.g. Ubuntu 19.04)


GabrielMachine:
I will push these changes to the other branch

PR #1636, opened 2019-11-10
[Multiview] Enhance L2 Rotation Averaging with Spectra library

### LINKED ISSUE #937: why use ROTATION_AVERAGING_L2 as default in global rotation computation
when the image count is large(more than 5000),i found that the global rotation computation use ROTATION_AVERAGING_L2 is very slow,but use ROTATION_AVERAGING_L1 is fast.
so my question is :why use ROTATION_AVERAGING_L2 as default in global rotation computation?is there some weaknesses for ROTATION_AVERAGING_L1?

rjanvier:
This PR integrate the Spectra library as a third party dependency into the build tree and take advantage of its features in the L2 RA solver. On some medium sized datasets, speed increase of L2RA is up to 10x. I haven't check for memory usage but it should be highly reduced because we use sparse matrices everywhere.
This hopefully definitely close #937


pmoulon:
Tested on my machine `openMVG_test_rotation_averaging (Failed)`

rjanvier:
The paramameters that control the convergence speed should be lowered. (see `nvec` parameter in the documentation
https://spectralib.org/doc/classspectra_1_1symeigssolver). Initially I use a higher value (7) than the recommended one (2*3 = 6). It worked on my computer but it failed on travis (and on your computer). The recommended value should work on your computer but it leeds to instabilities on travis (see this build https://travis-ci.org/rjanvier/openMVG/builds/612069353) so I lowered it (5) to make tests pass. I don't know if it's really rigorous...

PR #1653, opened 2019-12-11
fix histogram.hpp no such file or directory

belosthomas:
This commit fix this error : 
`/usr/local/include/openMVG/robust_estimation/robust_estimator_ACRansac.hpp:50:10: erreur fatale: third_party/histogram/histogram.hpp : Aucun fichier ou dossier de ce type`


pmoulon:
Build is failing for Linux and Windows. 
https://ci.appveyor.com/project/pmoulon/openmvg/builds/29466288
https://travis-ci.org/openMVG/openMVG/builds/623647281

As I said, the CMake Interface would be a better fix. I also think that on my system I can include this file without compiling error.

Can you share here your CMakeLists.txt to be sure you are including OpenMVG in the right way for your project?

PR #1665, opened 2020-01-05
[Matching] Add HNSWL2 matching with HNSWLib

rjanvier:
This PR Add ability to perform matching using Hierarchical Navigable Small World graphs as described in Yu. A. Malkov, D. A. Yashunin, [Efficient and robust approximate nearest neighbor search using Hierarchical Navigable Small World graphs](https://arxiv.org/abs/1603.09320) . 

- The implementation leverage [HNSWLIB](https://github.com/nmslib/hnswlib).
- The new HNSWL2 matcher is available both for float (Akaze) and uint8 (SIFT) descriptors in main_ComputeMatches.

## Speed
Here we tune the algorithm for a good balance between speed and precision.
In a general matter Indexing is slower than FLANN but matching is faster.  Exhaustive matching on big dataset would be faster.
Matching  43 images (903 pairs with way too much SIFT descriptors) would take 
285s with HNSW
344s with FLANN
299s with FASTCASCADE
_Please note that the way HNSW is tuned, there is a slight decrease in memory consumption vs. FLANN_

## Performance
On the Precision/Recall side, it behaves closer to BruteForce matcher than FLANN or CASCADEHASHING. It will produce less matches than the two latter options but on my opinion less outliers. For example on a given pair:
BF matches: 4508
HNSW matches: 4504 --> intersection BF/HNSW 4478
FLANN matches: 4638 --> intersection BF/FLANN 4200
CASCADE matches: 4568 --> intersection BF/CASCADE 3532

More benchmarks to come but reading the work of @vadixidav on the implementation of HNSW in Rust was very inspiring. Thanks.

### DIFF
--- a/src/software/SfM/main_ComputeMatches.cpp
+++ b/src/software/SfM/main_ComputeMatches.cpp
@@ -126,6 +126,7 @@ int main(int argc, char **argv)
       << "  AUTO: auto choice from regions type,\n"
       << "  For Scalar based regions descriptor:\n"
       << "    BRUTEFORCEL2: L2 BruteForce matching,\n"
+      << "    HNSWL2: L2 Approximate Matching with Hierarchical Navigable Small World graphs (float only),\n"
       << "    ANNL2: L2 Approximate Nearest Neighbor matching,\n"
       << "    CASCADEHASHINGL2: L2 Cascade Hashing matching.\n"
       << "    FASTCASCADEHASHINGL2: (default)\n"
@@ -333,6 +334,12 @@ int main(int argc, char **argv)
     {
       std::cout << "Using BRUTE_FORCE_HAMMING matcher" << std::endl;
       collectionMatcher.reset(new Matcher_Regions(fDistRatio, BRUTE_FORCE_HAMMING));
+    }
+    else
+    if (sNearestMatchingMethod == "HNSWL2")
+    {
+      std::cout << "Using HNSWL2 matcher" << std::endl;
+      collectionMatcher.reset(new Matcher_Regions(fDistRatio, HNSW_L2));
     }
     else
     if (sNearestMatchingMethod == "ANNL2")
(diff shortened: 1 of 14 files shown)


src/software/SfM/main_ComputeMatches.cpp
```diff
@@ -126,6 +126,7 @@ int main(int argc, char **argv)
       << "  AUTO: auto choice from regions type,\n"
       << "  For Scalar based regions descriptor:\n"
       << "    BRUTEFORCEL2: L2 BruteForce matching,\n"
+      << "    HNSWL2: L2 Approximate Matching with Hierarchical Navigable Small World graphs (float only),\n"
```
pmoulon: I'm removing float only since you have enabled uchar as a type too

src/third_party/CMakeLists.txt
```diff
@@ -96,11 +96,11 @@ if (DEFINED OpenMVG_USE_INTERNAL_EIGEN)
   add_subdirectory(eigen)
 endif()
 
-list(APPEND directories cmdLine histogram htmlDoc progress vectorGraphics)
+list(APPEND directories cmdLine histogram htmlDoc progress vectorGraphics hnswlib)
 foreach(inDirectory ${directories})
 install(
   DIRECTORY ./${inDirectory}
-  DESTINATION include/openMVG/third_party/
+  DESTINATION ${CMAKE_INSTALL_PREFIX}/include/openMVG/third_party/
```
pmoulon: Not sure we need this... Cmake should use it by default if CMAKE_INSTALL_PREFIX is defined.

src/software/SfM/main_ComputeMatches.cpp
rjanvier: yes was an artifact of my test implementation, sorry.

pmoulon:
Thx you @rjanvier 
PR integrated, feel free to review https://github.com/openMVG/openMVG/commit/17f1269ff44c886b665bff9b70582dab893c465f

src/software/SfM/main_ComputeMatches.cpp
pmoulon: No worries. I was just double-checking!

src/third_party/CMakeLists.txt
rjanvier: yes I just checked in the doc, you are right

Issue #1700, opened 2020-03-22
Clarification on identity pose


csparker247:
I've been working on importing pose priors from a custom camera rig, and I want to make sure that I've correctly propagated the rotation matrices for my cameras. The [camera documentation](https://openmvg.readthedocs.io/en/latest/openMVG/cameras/cameras/) doesn't really clarify the camera orientation w.r.t the world frame as far as I can tell, but I've seen in another issue that the OpenMVG camera system is Z forward and Y up. I interpret this to mean that a camera with identity rotation:

1. Faces in the Z+ direction
2. The top of the image from that camera is oriented in the Y+ direction
3. The horizontal axis of the image from that camera moves from left: X+ to right: X-

An identity camera frustum would essentially look like this, where the red face is the top of the frustum:
![default_frustum00](https://user-images.githubusercontent.com/1434526/77252465-5fce4000-6c2a-11ea-9018-9729b4679c2e.png)

In my tests, it appears as though rotation is also left-handed, though I'm admittedly not convinced of this point.

Any help or clarification on this would be useful! Thanks!

pmoulon:
Hello,

When you are trying to import your own camera pose I often recommend using https://github.com/openMVG/openMVG/blob/develop/src/software/SfM/export/main_ExportCameraFrustums.cpp in order to see if your cameras are correctly oriented.

Feel free to make our export tool better for you and the community.

pmoulon:
As you said Y should point up, Z Front.
You can also see here some synthetic dataset we are using for many unit test https://github.com/openMVG/openMVG/blob/develop/src/openMVG/multiview/test_data_sets.cpp#L82

csparker247:
Thanks! I may open a PR that adds colors to the exported frustum. That would have been very helpful when I was debugging my rotations.

pmoulon:
Thank you!
I was also thinking to add a mode to export X,Y,Z unit colored lines using the `edge` element from PLY but I never made it.

csparker247:
Sounds good! I've added that to for discussion in the PR I just opened. I think that closes this issue out. Thanks!

PR #1702, opened 2020-03-24
(filters_frustum) Add colors to exported ply

csparker247:
This colors the faces of the frustums exported by `ExportCameraFrustums`. This is meant to be an aid to help with orientating poses visually. 

This PR is currently just basic implementation in order to begin discussing details.

Questions/Thoughts:

1.  Should colors be optional?
2. Is there a better/preferred way to define color constants? I haven't really looked into the other colorization features in openMVG yet. Is `image::RGBColor` a good const class? Does it have an `ostream` operator?
3. I used colors generated by MeshLab. Do we want to tweak these?
4. What will it take to [add axis lines](https://github.com/openMVG/openMVG/issues/1700#issuecomment-602950501)? By this, I think you mean the camera's axes of rotation, centered on the pinhole.

### DIFF
--- a/src/openMVG/sfm/sfm_data_filters_frustum.cpp
+++ b/src/openMVG/sfm/sfm_data_filters_frustum.cpp
@@ -10,6 +10,7 @@
 
 #include "openMVG/cameras/Camera_Pinhole.hpp"
 #include "openMVG/geometry/pose3.hpp"
+#include "openMVG/image/pixel_types.hpp"
 #include "openMVG/sfm/sfm_data.hpp"
 #include "openMVG/stl/stl.hpp"
 
@@ -19,19 +20,14 @@
 #include <iomanip>
 #include <iterator>
 
-static const std::string COLOR_BOTTOM = "107 118 178";
-static const std::string COLOR_LEFT = "112 178 107";
-static const std::string COLOR_TOP = "178 107 124";
-static const std::string COLOR_RIGHT = "178 175 107";
-static const std::string COLOR_FAR = "164 107 178";
-static const std::string COLOR_NEAR = "223 196 230";
-
 namespace openMVG {
 namespace sfm {
 
 using namespace openMVG::cameras;
 using namespace openMVG::geometry;
 using namespace openMVG::geometry::halfPlane;
+
+std::string RgbPLYString(const image::RGBColor& color);
 
 // Constructor
 Frustum_Filter::Frustum_Filter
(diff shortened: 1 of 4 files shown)


pmoulon:
> Should colors be optional?
- Could be.
- I'm ok with default could having one top color different than the side face.

> Is there a better/preferred way to define color constants? I haven't really looked into the other colorization features in openMVG yet. Is image::RGBColor a good const class? Does it have an ostream operator?
- Yes there is color constant defined in image namespace. You are right it could be nice to define those with ostream iterator or having a std::to_string(RGBColor) working on it. The stream operator would have to match the `PLY` requirement `R G B`

> I used colors generated by MeshLab. Do we want to tweak these?
- Your initial suggestion was nice (top face with one color, and all the other with another fixed color)

> What will it take to add axis lines? By this, I think you mean the camera's axes of rotation, centered on the pinhole.
- It would consist in adding Edge to the PLY files (see here -> http://paulbourke.net/dataformats/ply/)
- Edges in PLY are linking some vertex with a Line (I don't know if it is supported by all 3D viewer) but it is a nice way to display 3D lines. I would like to use this feature for a while to display the visibility graph of the sfm_data scene but time is lacking... We could experiment here and see how easy to use it is.

csparker247:
Since there's already an `ostream` operator for `Rgb`, I'm not sure we can go that route unless you don't mind changing the format of that function from `{r, g, b}` to `r g b`. It seems like it's probably best to not mess with it. I'll look into alternatives.

csparker247:
Updated with: 
1. Uses `image::RGBColor` for colors
2. Adds RGB->PLY Color string function within the Filter_Frustum compilation unit.
3. Only the top face is colored red. Others are middle gray.
4. Colorization optional. Updated `ExportCameraFrustums` accordingly.

Still going to look into the axis lines using PLY Edges.

pmoulon:
Looking good for me as it is.
I tried and I saw that PLY EDGES are not supported by all the 3DViewer, so let's keep what we have.
Or we could add the axis here https://github.com/openMVG/openMVG/tree/develop/src/software/SfMWebGLViewer

I think the best for the future would be to write our own sfm_data viewer (3D scene) and editor (to edit ID and camera model).

src/openMVG/sfm/sfm_data_filters_frustum.cpp
```diff
@@ -181,26 +192,28 @@ const
 
   // Export frustums faces
   IndexT count = 0;
+  std::string color_top_face = colorize ? RgbPLYString(image::RED) : "";
```
pmoulon, line 195: Nit: consider using `const` if we know we not gonna modify it later

pmoulon:
Thx you @csparker247 
I cannot wait to work with you on your next contribution.

csparker247:
No problem! Glad to help!

Issue #1555, opened 2019-07-04
Export to openMVS does not support grayscale images


pivpiv:
I am trying to export a sfm project to OpenMVS, but it seems that the code does not handle black and white images. The program crash with the following output : 
"Catched a memory error in the image conversion. Please consider to use less threads ([-n|--numThreads])."
I tried to set the number of threads to 1, but the result is the same. 
When debugging the program it appears that the input images cannot be read because the depth is equal to 1. 
Other program, such as openMVG_main_ExportUndistortedImages seems to handle black and white images.

pmoulon:
Added your request as a feature for the next OpenMVG version.
Let's try to work on this in the coming weeks.

PR #1709, opened 2020-04-05
fix: #1555 Export to openMVS does not support grayscale images

junlinp:
Add some code for reading the grayscale with the openMVG Image Class.


pmoulon:
Thank you @junlinp

PR #1740, opened 2020-05-14
uf_tree indexing fix

greenlet:


greenlet:
I measured performance like in the code for the `sceaux_global_reconstruction` set:
1. `develop` baseline gives 3.5 sec
2. `issue-1737-experimental` with `std::map<indexedFeaturePair, uint32_t> image_feature_index` instead of `std::set<indexedFeaturePair> allFeatures` plus `flat_pair_map<indexedFeaturePair, uint32_t> map_node_to_index` gives 4.2 sec. So `std::map` didn't work out
3. Current `issue-1737` with `std::unordered_set<indexedFeaturePair, image_feature_hash> allFeatures` and  `flat_pair_map<indexedFeaturePair, uint32_t` gives 2.8 sec. Also `uf_tree` indexing fixed
https://github.com/openMVG/openMVG/pull/1740/files#diff-cb5e7e1f5559c0a219a2f49376872a98R151
https://github.com/openMVG/openMVG/pull/1740/files#diff-cb5e7e1f5559c0a219a2f49376872a98R199

src/openMVG/tracks/tracks.hpp
```diff
@@ -71,7 +78,14 @@ struct TracksBuilder
   {
     // 1. We need to know how much single set we will have.
     //   i.e each set is made of a tuple : (imageIndex, featureIndex)
-    std::set<indexedFeaturePair> allFeatures;
+    std::unordered_set<indexedFeaturePair, image_feature_hash> allFeatures;
```
greenlet, line 81: Is it possible to have two similar `indexedFeaturePair` at all?

pmoulon: We should not since we are using a set

pmoulon:
I made some benchmark on my machine and I found actually that your version is a bit slower.

Here the results on various datasets.

|Dataset|Time Old(s) |Time New(s)|%|
|--|--|--|--|
|chambord|**12.3622**|15.4362|.8|
|Dataset_Saint_Lys/Eglise|**5.17**|5.88403|.87|
|2017/LaPush/RubyBeach| **0.518462**| 0.658825|.78|
|2016/Eglise_Cognat|**17.0268**|20.2247|.85|

So on my machine, your code is in average slower (20%)

greenlet:
> I made some benchmark on my machine and I found actually that your version is a bit slower.
> 
> Here the results on various datasets.
> 
> Dataset	Time Old(s)	Time New(s)	%
> chambord	**12.3622**	15.4362	.8
> Dataset_Saint_Lys/Eglise	**5.17**	5.88403	.87
> 2017/LaPush/RubyBeach	**0.518462**	0.658825	.78
> 2016/Eglise_Cognat	**17.0268**	20.2247	.85
> So on my machine, your code is in average slower (20%)

What targets your test included? Please note that in current version there is a cycle which should be removed before testing `main_exportTracks`:
https://github.com/openMVG/openMVG/pull/1740/files#diff-278208fdb61f1a04029842126484e631R129-R137

pmoulon:
It could due to the compiler, I will test on another machine to see the impact.
Good point: Memory consumption was identical for both algorithms

PR #1741, opened 2020-05-14
Less work for image feature tracking

greenlet:
FYI. There's no major improvements, just less work and less code as flat_pair_map could be removed

### DIFF
--- a/src/openMVG/tracks/tracks.hpp
+++ b/src/openMVG/tracks/tracks.hpp
@@ -63,7 +63,7 @@ struct TracksBuilder
 {
   using indexedFeaturePair = std::pair<uint32_t, uint32_t>;
 
-  flat_pair_map<indexedFeaturePair, uint32_t> map_node_to_index;
+  std::map<indexedFeaturePair, uint32_t> image_feature_index;
   UnionFind uf_tree;
 
   /// Build tracks for a given series of pairWise matches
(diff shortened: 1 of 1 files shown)


pmoulon:
Did you compare speed (before/after)?
i would assume that inserting in a map is slower than inserting in a vector and sorting.

greenlet:
> Did you compare speed (before/after)?
> i would assume that inserting in a map is slower than inserting in a vector and sorting.

I haven't made measurements. But inserting in a map has the same complexity as inserting into set
```std::set<indexedFeaturePair> allFeatures;```
`std::map` and `std::set` are both sorted structures

pmoulon:
Let's try to bench on many scenes and then consider to merge depending on the timing results.

You can add a timer here on this binary and quit early to benchmark (Running X times in a loop could also help to have more accurate timing) https://github.com/openMVG/openMVG/blob/develop/src/software/SfM/export/main_exportTracks.cpp#L114

pmoulon:
The key for speed is to avoid repetitive memory allocation, using the flat_map allow to have already the memory and insertion gonna be quick. Using a set will guarantee quick insertion but will make repetitive memory allocation.

Let's benchmark it and see ;-)

greenlet:
Ok. I can do benchmark later

PR #1760, opened 2020-06-19
Enable the 'NEW' mode of CMP0100

cwgthornton:
Without this, new versions of cmake do not run automoc on .hh files,
generating linker errors for the ui tools.


pmoulon:
Thank you @cwgthornton

PR #1762, opened 2020-06-24
[SfM/export] Fix bug when exporting sfm_data with unknown intrinsics …

TongLing916:
Fix bug when executing `openMVG_main_openMVG2openMVS` using a `sfm_data.json` which contains some views with unknown intrinsic parameters (i.e, `"id_intrinsic": 4294967295`). 

Tested dataset: [Photo Tourism](http://phototour.cs.washington.edu/datasets/)

### DIFF
--- a/src/software/SfM/export/main_openMVG2openMVS.cpp
+++ b/src/software/SfM/export/main_openMVG2openMVS.cpp
@@ -86,20 +86,25 @@ bool exportToOpenMVS(
   scene.images.reserve(nViews);
   for (const auto& view : sfm_data.GetViews())
   {
-    map_view[view.first] = scene.images.size();
-    MVS::Interface::Image image;
+    ++my_progress_bar;
+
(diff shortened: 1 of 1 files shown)


src/software/SfM/export/main_openMVG2openMVS.cpp
```diff
@@ -86,6 +86,11 @@ bool exportToOpenMVS(
   scene.images.reserve(nViews);
   for (const auto& view : sfm_data.GetViews())
   {
+    if (map_intrinsic.count(view.second->id_intrinsic) == 0)
```
pmoulon, line 89: I would recommend to use `IsPoseAndIntrinsicDefined` : https://github.com/openMVG/openMVG/blob/develop/src/openMVG/sfm/sfm_data.hpp#L61

TongLing916: Thanks for your advice. 

Fixed, and removed some redundant and ambiguous codes (in my opinion).

PR #1769, opened 2020-07-08
Catches template type incompatibility early in HierarchicalNSW::Build

flosincapite:
HNSWmatcher will throw an error if constructed when HNSWmetric is nullptr.

### DIFF
--- a/src/openMVG/matching/matcher_hnsw.hpp
+++ b/src/openMVG/matching/matcher_hnsw.hpp
@@ -51,10 +51,18 @@ class HNSWMatcher: public ArrayMatcher<Scalar, Metric>
     int dimension
   ) override
   {
+    HNSWmetric.reset(nullptr);
+    HNSWmatcher.reset(nullptr);
+    
     if (nbRows < 1)
-    {
-      HNSWmetric.reset(nullptr);
-      HNSWmatcher.reset(nullptr);  
+      return false;
+      
+    // Returns early if DistanceType is not supported.
+    // This avoids the case where HNSWmetric is null, which causes an error when 
+    // HNSWmatcher initializes.
+    if (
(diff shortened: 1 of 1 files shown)


src/openMVG/matching/matcher_hnsw.hpp
```diff
@@ -51,10 +51,18 @@ class HNSWMatcher: public ArrayMatcher<Scalar, Metric>
...
-      HNSWmetric.reset(nullptr);
-      HNSWmatcher.reset(nullptr);  
+      return false;
+      
+    // Returns early if DistanceType is not supported.
+    // This avoids the case where HNSWmetric is null, which causes an error when 
+    // HNSWmatcher initializes.
+    if (
+        typeid(DistanceType) != typeid(int) &&
+        typeid(DistanceType) != typeid(float)) {
```
simogasp, line 65: maybe remove the else branch below and put here the
`std::cerr << "HNSW matcher: this type of distance is not handled Yet" << std::endl;`

flosincapite: Cool, taken care of.

Still seems like there's more that could be done. What do you think of using explicit template instantiations for int and float? That would cause compile errors instead of this soft error, so it changes the interface. What do you all think?

flosincapite:
Hi! Apologies for the delay. Is this ready to merge or would it make sense to explore alternative options?

rjanvier:
Hi! No problem. Thanks a lot for your PR :). I will finish and try to merge #1771 before focusing on this one with the team.

pmoulon:
Has been merged via https://github.com/openMVG/openMVG/commit/188c674c69f02b9cd69ffaa9152ca14a150f5cb2

PR #1770, opened 2020-07-08
[Camera] Change the way Brown and Fisheye models are serialized

rjanvier:
Currently used method add an additional "value0" field to Brown and Fisheye models that add aditional complexity to the parsing of json/xml by external process/libs.

it should be integrated into a future 2.0 release as it break compatibility


pmoulon:
Has been merged by https://github.com/openMVG/openMVG/commit/16557e359a7715f42ab33f3153f02c2d93fabc2f

PR #1771, opened 2020-07-08
[Matching] Matching kernels for HNSW

rjanvier:
- Fix HNSW performance drop on Windows systems where `__SSE__` is not defined for `/arch: SSE2` and subs. (should be ported upstream)
- the current L2 FLOAT AVX2 kernel is in fact a FLOAT **AVX** kernel. this PR modified this kernel and all depending functions/tests. Furthermore, `metric_avx2.hpp` is renamed `metric_simd.hpp`.
- Add a L1 metric and also a HNSW L1 matcher (taking inspiration from recent papers by [Fabio Bellavia](https://www.researchgate.net/publication/338764333_Is_there_anything_new_to_say_about_SIFT_matching) ) for `uint8` features. L1 matching is always faster vs. L2 matcher with no or maybe very minor accuracy drop (in fact accuracy could be **better** but it needs to be carefully quantified)
- Add an HNSW HAMMING matcher
- New default matchers in `main_ComputeMatches`: HNSWL1 for uint8 descriptors (SIFT, LIOP), HNSWL2 for float based descriptors and HNSWHAMMING for binary descriptors

Timings:

Chapelle JB - 10K pairs / 50 Kfeats per images on average (on a 24 threads low frequency / max. AVX Linux computer):
- HNSWL1 approx. 1800 s. (Multiple runs with best at 1746 s.)
- HNSWL2 approx. 2000 s. (Multiple runs with best at 2028 s.) 
- CASCADE 2526 s. (One run)

Cascade hashing is faster on AVX2 computers.


src/software/SfM/main_ComputeMatches.cpp
```diff
@@ -127,13 +127,16 @@ int main(int argc, char **argv)
...
+      << "      taylored for quantized and histogram based descriptors (e.g uint8 RootSIFT)\n"
       << "    ANNL2: L2 Approximate Nearest Neighbor matching,\n"
       << "    CASCADEHASHINGL2: L2 Cascade Hashing matching.\n"
       << "    FASTCASCADEHASHINGL2: (default)\n"
       << "      L2 Cascade Hashing with precomputed hashed regions\n"
       << "     (faster than CASCADEHASHINGL2 but use more memory).\n"
       << "  For Binary based descriptor:\n"
-      << "    BRUTEFORCEHAMMING: BruteForce Hamming matching.\n"
+      << "    BRUTEFORCEHAMMING: BruteForce Hamming matching,\n"
+      << "    HNSWHAMMING: Hamming pproximate Matching with Hierarchical Navigable Small World graphs\n"
```
pmoulon, line 139: typo `pproximate`

src/openMVG/matching/metric_hnsw.hpp
```diff
@@ -0,0 +1,197 @@
...
+{
+  size_t size = *(reinterpret_cast<const size_t*>(qty_ptr));
+  size = size >> 2;
+  int result = 0;
+  uint8_t *a = (uint8_t *)(pVect1); // discard const
+  uint8_t *b = (uint8_t *)(pVect2); // discard const
+
+
+  // Process 4 items for each loop for efficiency.
+  for (size_t i = 0; i < size; i++) {
```
pmoulon, line 75: Does Eigen absis vectorized and could be used here?

src/openMVG/matching/matcher_hnsw.hpp
```diff
@@ -103,9 +137,9 @@ class HNSWMatcher: public ArrayMatcher<Scalar, Metric>
     DistanceType * distance
   ) override
   {
-    if (HNSWmatcher.get() == nullptr)
+    if (HNSW_matcher_.get() == nullptr)
```
pmoulon, line 140: if we have smart pointer doing `if (!HNSW_matcher_)` should be ok

src/openMVG/matching/metric_hnsw.hpp
rjanvier: you mean putting this in a Eigen::Map and using the dedicated abs? the issue here is that the size is only known at runtime, it's not sure Eigen will be able to unroll it properly, I'll have to test.

src/openMVG/matching/matcher_hnsw.hpp
```diff
@@ -103,9 +135,10 @@ class HNSWMatcher: public ArrayMatcher<Scalar, Metric>
     DistanceType * distance
   ) override
   {
-    if (HNSWmatcher.get() == nullptr)
+    if (!HNSW_matcher_)
       return false;
-    auto result = HNSWmatcher->searchKnn(query, 1).top();
+    HNSW_matcher_->setEf(16); // it could probabely by lowered in this case
```
pmoulon, line 140: typo  "probabely"

pmoulon:
Has been merged by https://github.com/openMVG/openMVG/commit/c63f0b717c09cf97105bc225d515ddb7fed6e9ee

PR #1773, opened 2020-07-22
(filters_frustum) Fix top face not being colored correctly

csparker247:
This fixes a bug introduced in #1702 in the frustum PLY exporter where the wrong face was being colored as the top face. This patch correctly colors the top face of both the infinite and truncated frustums.

Using the Sceaux Castle dataset + global SfM, this is the new output of `openMVG_main_ExportCameraFrustums`:

frustum-color-fixed-01

And this is the same for a truncated frustum:

frustum-color-fixed-00

### DIFF
--- a/src/openMVG/sfm/sfm_data_filters_frustum.cpp
+++ b/src/openMVG/sfm/sfm_data_filters_frustum.cpp
@@ -199,9 +199,9 @@ const
   {
     if (it->second.isInfinite()) // infinite frustum: drawn normalized cone: 4 faces
     {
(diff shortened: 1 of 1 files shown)


SBCV:
Side note: If you want a bit more sophisticated visualization of the camera frustums, you could use [this blender addon ](https://github.com/SBCV/Blender-Addon-Photogrammetry-Importer) to import the OpenMVG reconstruction and scale the cameras accordingly.

pmoulon:
Thx you @csparker247

PR #1812, opened 2020-10-13
improve OpenMVS interface

cdcseacave:
Hi Pierre,

I just corrected a (relatively) small bug in OpenMVS that has implications in the interface data too. The problem is that the interface remains the same but the interpretation of the data changes slightly, and best would be to slightly adjust your exporter to remove this potential problem.

In short the pixel center in OpenMVS is assumed to be the integer coordinates and not with 0.5. This implies that the principal point has 0.5 subtracted, which make the original normalization scheme slightly off (the one now existing in your exporter). So best would be to completely remove the normalization process and just set the image resolution. Pls adjust in case OpenMVG has a different convention for the pixel center.

Thanks!


pmoulon:
Thank you @cdcseacave

Issue #1788, opened 2020-08-23
How to create new Regions object and also modify feature coordinates?


meder411:
I am putting together what hopefully will be a PR with some stuff for spherical image feature detection. However, I'm running into the issue of copying features from multiple images, each stored in their own `Regions` instance, to a single `Regions` instance. When doing this, I also want to modify the point coordinates (x, y) of the features. (If it helps, [here](https://github.com/meder411/openMVG/blob/develop/src/openMVG/spherical/tangent_images.cpp#L308) is where I'm trying to do this--hopefully the comments make it clear what's going on).

Basically I'm trying to do 2 things:

(1) Create a new `Regions` instance and fill it with copied data from a vector of other `Regions`

(2) Change the coordinates of each feature within the `Regions` object.

I have been able to handle (1) via the `EmptyClone()` function, but that feels like a hack. (2) doesn't seem possible from the abstract [`Regions`](https://github.com/openMVG/openMVG/blob/develop/src/openMVG/features/regions.hpp) class, but  both [`Scalar_Regions`](https://github.com/openMVG/openMVG/blob/160643be515007580086650f2ae7f1a42d32e9fb/src/openMVG/features/scalar_regions.hpp) and [`Binary_Regions`](https://github.com/openMVG/openMVG/blob/160643be515007580086650f2ae7f1a42d32e9fb/src/openMVG/features/binary_regions.hpp) have mutable accessors to the [`PointFeatures`](https://github.com/openMVG/openMVG/blob/develop/src/openMVG/features/feature.hpp) stored within. In order to be compatible with any type of `Regions` instance, I think I need my function to accept this base class, but there's no way to modify the feature coordinates through the `Regions` class.

How can I go about doing this while maintaining the flexibility of accepting any sub-type of `Regions`?

pmoulon:
Hello @meder411 
Thank you for asking for help and advice on this.
By default, the abstraction has been done to make Regions not mutable and then use the child classes to populate accordingly the descriptor and feature that could be a template.

- I would advise you either to use a template function on the Regions Type and casting before, so you would have access to the mutable accessor.

- Else we could hack the API to add a generic function to the `Regions` class (UpdatePosition(x,y)) if we assume that all Regions type would have at least a position.

Does that make sense?

It is hard to find a very extensible API and our design is not perhaps the best to allow this kind of action.
[...]

PR #1814, opened 2020-10-17
Tangent images 

meder411:
* Added class for creating and working with tangent images
* Class exposes 2 primary functions:
    1. Creating tangent images (with associated valid-region masks) from an input equirectangular image
    2. Converting pixel coordinates on a given tangent image back to pixel coordinates on the input equirectangular image

The first function is, obviously, to create the tangent image set. It also produces masks designating which part of the tangent image maps directly to the face of the icosahedron. The valid regions are given the value 255 and the invalid (padding) region is 0. These masks should be provided via the `mask` parameter to sparse feature detection functions to avoid redundant feature computation across the tangent image set.

The second function is for converting the detected feature coordinates from the tangent image grid back to the equirectangular grid. The benefits of tangent images are their low-distortion representation. As such we ought to compute the descriptor on the tangent image, but we often want to then transfer those keypoints and descriptors back to the original equirectangular image before running our correspondence search with a spherical camera model. 

**Note:** currently, this PR does not have a way to actually update the coordinate of any computed feature. I'm not entirely sure how to do this (see my [question here](https://github.com/openMVG/openMVG/issues/1788)). I welcome any guidance or help creating a `UpdateFeatureCoordinates()` function.

A unit test is provided demonstrating how to create the tangent images and masks. I will add a second unit test demonstrating the tangent-image-to-equirectangular conversion along with feature matching once I have found a way to perform the coordinate update for the features.


meder411:
CodeFactor doesn't seem to like my usage of `dynamic_cast` to disambiguate the `Regions` object in [`ComputeFeaturesOnTangentImages`](https://github.com/openMVG/openMVG/blob/84f0026a89b83874fc5bdb3ea524002e5ae94901/src/openMVG/spherical/tangent_images.cpp#L409). Any alternative solutions come to mind?

meder411:
I exposes a few additional functions (mainly coordinate conversion operations) and fixed a bug with the coordinate conversions used for image description. I also added a bunch of unit tests for different components as well as a image resampling from tangent images back to an equirectangular image.

Probably should pause on merging the PR for a bit though as there seems to be something odd with the SIFT detection on the tangent images themselves. In the paper, I used OpenCV's SIFT via Python for detection and didn't have these issues. All the coordinate conversions match up fine, but simply running the SIFT_image_describer on a tangent images in OpenMVG looks off:

image

I'm working on figuring out what's up.

Issue #1529, opened 2019-05-23
Cannot link libraries when running as a third-party libary using Makefile


BrennoCaldato:
I'm having some problems while using openmvg as a third-party library compiled by a Makefile. 

I'm using the following Makefile:

```
include ../Makefile.common

LIBS_OPENMVG = -lopenMVG_image  -lopenMVG_features -lopenMVG_sfm  -lopenMVG_matching_image_collection \
			   -lopenMVG_system -lopenMVG_exif  -lopenMVG_stlplus -lvlsift 


FLAGS+=-I/usr/local/include/openMVG/
FLAGS+=-I/usr/local/include/openMVG_dependencies/
FLAGS+=-I/usr/local/include/openMVG_dependencies/cereal/include/



TST_IMG_LOCALIZER:=tests/test_image_localizer.o image_localizer.o

.PHONY: clean all
	
all: test_image_localizer

test_image_localizer: $(TST_IMG_LOCALIZER) 
	$(CXX) -o tests/test_image_localizer $(TST_IMG_LOCALIZER) -pedantic $(FLAGS) -I$(INC) $(LIBS) $(LIBS_OPENMVG) -lgtest

test:
	tests/test_image_localizer

clean:
	$(RM) *.o *.d *.so tests/*.o tests/*.d $(TST_IMG_LOCALIZER) tests/test_image_localizer
	find . -type f \( -name '*.gcda' -o -name '*.gcno' \) -delete
```

And get the output:

```
g++ -MM -I/home/rachillesf/repo-migration/image_localization -I/usr/local/include -I/usr/local/include/cm -std=c++11 -Wall -Wno-variadic-macros -g -fPIC -DCM_BUILD_ENV='"master"' -DCM_VERSION='""' -DCM_GIT_HASH='"b5b9aa1"' -D__STDC_CONSTANT_MACROS -D__STDC_LIMIT_MACROS -fopenmp -O3 -Ofast -I/usr/local/include/openMVG/ -I/usr/local/include/openMVG_dependencies/ -I/usr/local/include/openMVG_dependencies/cereal/include/ -fopenmp -I/home/rachillesf/repo-migration/image_localization image_localizer.cpp > image_localizer.d
[...]

PR #1831, opened 2020-12-19
Fix missing ceres libaries

bartoszek:
Had same problem as in #1529 linking target `openMVG_main_ComputeStructureFromKnownPoses` misses `ceres` in linker call.
```
[ 90%] Linking CXX executable ../../Linux-x86_64-RELEASE/openMVG_main_ComputeStructureFromKnownPoses
/usr/bin/ld: CMakeFiles/openMVG_main_ComputeStructureFromKnownPoses.dir/main_ComputeStructureFromKnownPoses.cpp.o: undefined reference to symbol '_ZN5ceres41IsSparseLinearAlgebraLibraryTypeAvailableENS_30SparseLinearAlgebraLibraryTypeE'
/usr/bin/ld: /usr/lib/libceres.so.2: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
make[2]: *** [software/SfM/CMakeFiles/openMVG_main_ComputeStructureFromKnownPoses.dir/build.make:112: Linux-x86_64-RELEASE/openMVG_main_ComputeStructureFromKnownPoses] Error 1
```
Looks like `STLPLUS_LIBRARY` requires `CERES_LIBRARIES`  when linked with shared `libceres`.

### DIFF
--- a/src/software/SfM/CMakeLists.txt
+++ b/src/software/SfM/CMakeLists.txt
@@ -161,6 +161,7 @@ target_link_libraries(openMVG_main_ComputeStructureFromKnownPoses
     openMVG_features
     openMVG_sfm
     ${STLPLUS_LIBRARY}
+    ${CERES_LIBRARIES}
 )
 
 add_executable(openMVG_main_ComputeSfM_DataColor main_ComputeSfM_DataColor.cpp)


pmoulon:
Ceres library should come by transitivity from openmvg_sfm library.
Or the mistake is that I put it as PRIVATE here https://github.com/openMVG/openMVG/blob/develop/src/openMVG/sfm/CMakeLists.txt#L34 and it should not be.

Let's investigate a bit. Thank you for pointing this up!

pmoulon:
Fixed by https://github.com/openMVG/openMVG/commit/dded4e46e645a6bfe566e84c18a723f1aa47ceca

Issue #1852, opened 2021-02-01
[openSUSE Tumbleweed] Found some file went into /usr/lib, should it go to /usr/lib64 when building on 64 bit arch?


kevinsmia1939:
Hello,

I build openMVG package and found some file went into /usr/lib but shoud it go to /usr/lib64 when building on 64 bit architecture?

I am not sure, but could this line be the issue?
https://github.com/openMVG/openMVG/blob/5e98d504bb76ba2d1d07ae80ac2acb10b3d6f97d/src/CMakeLists.txt#L552

Would it be better to replace lib with ${CMAKE_INSTALL_LIBDIR}?

Also why is build share libs is off by default?
https://github.com/openMVG/openMVG/blob/5e98d504bb76ba2d1d07ae80ac2acb10b3d6f97d/src/CMakeLists.txt#L25

I was told by other Linux packager that share libs are prefered to static libs (.a .la) and told me I should delete them.
Let me know what you think.

pmoulon:
Hello @kevinsmia1939 ,
Thank you for your feedback.

You are right, we certainly not are using CMake at is full potential to export the lib at the right place.
Thank you for pointing this out.
-> Do you want to make a PR to fix this?

Regarding Shared libs...
-> I personally prefer to use static libs when developing, so I'm absolutely sure about which code is used in the binaries.

Feel free to activate the shared libs using cmake command line for your build `-OpenMVG_BUILD_SHARED:bool=ON` and to create your package.

PR #1855, opened 2021-02-06
Master

kevinsmia1939:
Use GNUInstallDirs for directories. Fix for this #1852.
I tried using openMVG pipeline to check if it work, it does work, but check to make sure.

Kind regards


pmoulon:
Thank you. I will take a look during the week and let you know. FYI when you run install there is a manifest file that trace all the file that have been install. By looking to it you can check if there is any file going in the wrong path.

kevinsmia1939:
> Thank you. I will take a look during the week and let you know. FYI when you run install there is a manifest file that trace all the file that have been install. By looking to it you can check if there is any file going in the wrong path.

Hello,

I check all the files, it went to the correct path.
I only apply gnuinstalldir to lib only, I left bin and include alone because I don't know if I will break them, but if you want me to cover them too let me know.

pmoulon:
Let's do step by step. Starting with the LIB is a good start

src/openMVG/CMakeLists.txt
```diff
@@ -1,3 +1,4 @@
+include(GNUInstallDirs)
```
pmoulon, line 1: I think we should do this include only once (only in the `primary` cmake the one in src/CMakeLists.txt)

pmoulon:
Solved by integrating #2148

PR #1871, opened 2021-03-20
[Feature] Anatomy Use L2 norm before clamping sift desc, even for root sift

rjanvier:
In rootsift, L1 norm. is used after the first round of L2 norm. + clamping, not before clamping.


pmoulon:
LGTM Thank you
image

PR #1877, opened 2021-03-27
Develop 2.0 build shared

pmoulon:
Enhance support of SHARED library build
- Work now for VStudio/VSCode
- Proposal to use Lemon as a header only library (CMake INTERFACE) -> Perhaps some more CMake cleaning there could be handy


RamadanAhmed:
Hi, 
Why not using cmake export headers and add export headers to all needed classes and free functions ?

pmoulon:
@RamadanAhmed  I prefer the non invasive solution for now.

PR #1890, opened 2021-04-30
small utility to convert matches from .bin to .txt or .txt to .bin

yocabon:


src/software/SfM/main_ConvertMatches.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+// Copyright (c) 2012, 2013, 2015 Pierre MOULON.
+
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+#include "openMVG/matching/indMatch.hpp"
+#include "openMVG/matching/indMatch_utils.hpp"
+#include "openMVG/matching/svg_matches.hpp"
```
pmoulon, line 11: not needed

src/software/SfM/main_ConvertMatches.cpp
```diff
@@ -0,0 +1,94 @@
...
+#include "openMVG/image/image_io.hpp"
```
pmoulon, line 12: not needed

pmoulon (changes requested):
Thank you for your contribution
I just advised some cleanup and directly using Load/Save function from the feature library files.
It will make the code even shorter and simpler.

yocabon (commented):
Done

pmoulon:
LGTM, thank you for your contribution

PR #1895, opened 2021-05-24
[SfM/export] adapt cubic images to the size of spherical images

nR3D:
Having a fixed size for cubic images can lead to a quality loss in case of large spherical images.

The size of cubic images can be better expressed as `spherical_image.Height()/2` (or alternatively `spherical_image.Width()/4`)


pmoulon:
Thank you for this contribution, I think it makes sense, could we find a tradeoff solution where we could set the resolution to auto or to a given size. For example you could have various resolution panorama and still export uniform resolution output (even if it will involve upsampling).
What do you think?

nR3D:
I could add a new command line parameter (`-s | --size-cubic` for instance). If you care about backward compatibility it can be set to `1024` by default, and if set to `-1` a variable resolution will be used.

pmoulon:
Your suggestion looks go.

Issue #1727, opened 2020-04-28
Mac Compiling issue


AndersonChang:
Hi all,

I tried to follow the steps :

1.  git clone --recursive https://github.com/openMVG/openMVG.git
2. mkdir openMVG_Build
3. cd openMVG_Build
4. cmake -DCMAKE_BUILD_TYPE=RELEASE . ../openMVG/src/
5. make install 

and then I got the following message:

[  0%] Building HTML documentation with Sphinx
Running Sphinx v2.4.0
WARNING: html_static_path entry '_static' does not exist
loading pickled environment... done
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 0 source files that are out of date
updating environment: 0 added, 0 changed, 0 removed
looking for now-outdated files... none found
no targets are out of date.
build succeeded, 1 warning.

The HTML pages are in htmlDoc.
[  0%] Built target doc
[ 14%] Built target lib_CoinUtils
[ 17%] Built target lib_Osi
[ 30%] Built target lib_clp
[ 31%] Built target lib_OsiClpSolver
[ 32%] Built target openMVG_stlplus
[ 34%] Built target openMVG_lemon
[ 34%] Built target main_svgSample
[ 35%] Built target openMVG_easyexif
[ 37%] Built target openMVG_fast
[ 37%] Built target openMVG_exif
[ 39%] Built target openMVG_features
[ 40%] Built target openMVG_numeric
[ 41%] Built target openMVG_image
[ 42%] Built target openMVG_linearProgramming
[ 48%] Built target openMVG_multiview
[ 49%] Built target openMVG_lInftyComputerVision
[ 51%] Built target openMVG_geometry
[ 52%] Built target openMVG_matching
[ 53%] Built target openMVG_kvld
[ 54%] Built target openMVG_matching_image_collection
[ 54%] Built target openMVG_multiview_test_data
[ 54%] Built target openMVG_robust_estimation
[ 54%] Built target openMVG_system
[ 60%] Built target openMVG_sfm
[ 60%] Linking CXX executable ../../Darwin-x86_64-RELEASE/openMVG_sample_cameras_undistoBrown
ld: library not found for -lgomp
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [Darwin-x86_64-RELEASE/openMVG_sample_cameras_undistoBrown] Error 1
[...]

PR #1908, opened 2021-06-22
builds on Mac OS 11 Big Sur AppleClang + OpenMP

rfabbri:
Builds with XCode 12.5.1
The problem was that cmake would proceed OK,
but there were build and link errors such as 'omp.h' not found,
and '-lomp' and '-lgomp' not found.

Reference fix: https://github.com/FluidSynth/fluidsynth/pull/919/commits/486e0f5e8f3a7b9d0bc6bb0587c9f7ec1dadabb1


rfabbri:
I get the same 'omp.h' not found when compiling with Xcode 13.3 under macOS Monterey 12.2.1 (march/2022).
I adapted this PR to the current `develop` branch, see PR  #2038  as a proposed fix to #1727 for the develop branch. The changes in this PR (that was based on the old develop_2.0 branch) may be useful for some other branch than `develop`.

PR #1932, opened 2021-09-10
Fix compilation error with eigen 3.4.0

aluaces:
Although I have been said that this issue is going to be eventually fixed in eigen 3.4.1, if I'm not mistaken the size of that AᵀA matrix is going to be always 9×9, so it's better being explicit.


pmoulon:
Thank you for noticing that.
Seems like that in fact we count factorize some code now.
I will try to look to it soon

pmoulon:
@aluaces I merged your update and factorized some code, can you let me know if everything looks good on your side too.

aluaces:
Everything is perfect, thanks!

PR #1934, opened 2021-09-13
Removed xfeatures2d and made it efficient with declaration

Cecilimon:
Hi,

I was working with the OpenCV version of `main_ComputeFeatures_OpenCV` and found that it still uses the SIFT implementation from xfeatures2d(OpenCV contrib).

So I changed the code to use the SIFT Implementation from features2d. and removed the inclusion of xfeatures2d.hpp

Also, I found that the code was creating the SIFT Object every time when Describing the Features. So I changed the code to create the SIFT Descriptor in a constructor and use the created SIFT Descriptor object while describing the feature

Plz take a look if it looks fine to you, and comments are very welcomed.

Thank you for your contribution to this great OpenSource.

### DIFF
--- a/src/software/SfM/main_ComputeFeatures_OpenCV.cpp
+++ b/src/software/SfM/main_ComputeFeatures_OpenCV.cpp
@@ -23,7 +23,7 @@
 #include <opencv2/opencv.hpp>
 #include "opencv2/core/eigen.hpp"
 #ifdef USE_OCVSIFT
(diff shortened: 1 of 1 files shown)


pmoulon:
LGTM, let me try on my side too in the next coming days and merge.
Thx you for your contribution and kind words towards the library.

pmoulon:
Thx @Cecilimon, update merged in develop_2.0 branch

PR #1942, opened 2021-10-11
Add near/far bounds to frustums visualization binary

kimoktm:
**Change:**
Add flags to `openMVG_main_ExportCameraFrustums` binary to allow the user to set near/far bounds of frustums. This makes it easier to visualize and debug frustums related functionalities such as frustums intersection. The functionality was already there, it just needed to be exposed from the cli.

Cli:
```sh
openMVG_main_ExportCameraFrustums -i sfm_data.bin -o frustums.ply -n NEAR_PLANE -f FAR_PLANE
```

**Examples:**
Before:
default

Custom near & far planes:
custom

Thanks @pmoulon for the great library 😊!!

### DIFF
--- a/src/software/SfM/export/main_ExportCameraFrustums.cpp
+++ b/src/software/SfM/export/main_ExportCameraFrustums.cpp
@@ -31,9 +31,13 @@ int main(int argc, char **argv)
 
   std::string sSfM_Data_Filename;
   std::string sOutFile = "";
+  double z_near = -1.;
+  double z_far = -1.;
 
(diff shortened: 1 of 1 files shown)


kimoktm:
LGTM C++ tests seems to be failing on `develop` because it runs out of memory. I checked the history when it started to fail [link](https://lgtm.com/projects/g/openMVG/openMVG/rev/a32ba1809269dd0ae0727db4e47faae5717f669d) - and nothing really changed in the C++ code. I think LGTM just need more RAM, what do you think @pmoulon?

PR #1961, opened 2021-11-07
Update BUILD.md

wolfiex:
We are already in the openMVG directory if following the instructions.


Janjs:
I think you would be at the openMVG_Build directory, not openMVG. Notice the last command (cd openMVG_Build) of the previous step.

pmoulon:
@wolfiex  Thx you for your PR but I think @Janjs is right, I will keep the documentation as it is.
Don't hesitate to make other PR to the project

PR #1964, opened 2021-11-15
VLAD Improvements

rjanvier:
- Install rules were not defined in CMake for main_ComputeVLAD binary
- BuildCodeBook method always use default parameter for codebooksize (128) this results in undefined behavior when setting another value with the command line parameter.


pmoulon:
Thank you @rjanvier!

PR #1970, opened 2021-11-29
[Multiview] Improve Nordberg p3p solver

rjanvier:
- Merge last changes from upstream thanks to @midjji
https://github.com/midjji/lambdatwist-p3p/commit/25eb8e2f5c2bde793657538d1a7dfd893109d706
and
https://github.com/midjji/lambdatwist-p3p/commit/c1c35975ccc3995de4a8fa029b29ebc8a76ff5a8
- Add const where its needed
- Remove unnecessary bearing normalization
- Add some refs / remove some useless refs
the solver is now faster and more robust.

### DIFF
--- a/src/openMVG/multiview/solver_resection_p3p_nordberg.cpp
+++ b/src/openMVG/multiview/solver_resection_p3p_nordberg.cpp
@@ -36,52 +36,52 @@ static void gauss_newton_refineL(Vec3 &L,
   // TODO(RJ:) I have hardcoded the number of iterations here, it's a template parameter in the original implementation
   for (int i = 0; i < 5; ++i)
   {
-    double l1 = L(0);
-    double l2 = L(1);
-    double l3 = L(2);
-    double r1 = l1 * l1 + l2 * l2 + b12 * l1 * l2 - a12;
-    double r2 = l1 * l1 + l3 * l3 + b13 * l1 * l3 - a13;
-    double r3 = l2 * l2 + l3 * l3 + b23 * l2 * l3 - a23;
+    const double l1 = L(0);
+    const double l2 = L(1);
+    const double l3 = L(2);
+    const double r1 = l1 * l1 + l2 * l2 + b12 * l1 * l2 - a12;
+    const double r2 = l1 * l1 + l3 * l3 + b13 * l1 * l3 - a13;
+    const double r3 = l2 * l2 + l3 * l3 + b23 * l2 * l3 - a23;
 
     if (std::abs(r1) + std::abs(r2) + std::abs(r3) < 1e-10)
       break;
 
-    double dr1dl1 = 2.0 * l1 + b12 * l2;
-    double dr1dl2 = 2.0 * l2 + b12 * l1;
+    const double dr1dl1 = 2.0 * l1 + b12 * l2;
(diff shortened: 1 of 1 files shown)


midjji:
The bearing normalization is required.

rjanvier:
Yep but we used to normalize twice before in openMVG. 1. We call the solver with normalized bearing and 2. I don't know why I added a normalization step into the solver... so I remove the one inside the solver because it's explicit in the function signature and documentation we should use bearing vector in input for lambdatwist.
new PR https://github.com/openMVG/openMVG/pull/1972 with good Eigen port of the computation of u1,u2,u3. in this PR, U coeffs make this one fail on test because Eigen indexing do not match CVL indexing (I think because of different row/column major storage)

midjji:
As long as the user knows they must normalize them before i suppose that is
fine. But its an easy mistake to make.

Both eigen and cvl use
Cols*row + col indexing. So that should not be it. Please send me the
xs,ys, and ill check it out. My randomly generated test cases does miss
very specific but non degen configurations.

Note, the optimizer cannot use const to improve performance. Constexpr does
let the compiler improve performance in general, but using it too much will
make you cry at compiletime.

Performancewize the most costly part by an extreme margin, as in more than
all of the rest is the heap allocation of the std::vector output. The real
trouble is if the local pool runs out and it has to go to the os to ask for
more. I recomend you make it an out parameter instead, and tell the user in
doc to prealloc it to size 4. The most common use case will be ransac, so
thats easy enough to do, but still very helpful, turning one alloc to a
single one, and making the p3p latency far more predictable.


Btw there is a version which outputs a quaternion instead of a rotation
matrix saving more time, and slightly improving num accuracy, if thats
interesting.

rjanvier:
Hey Mikael thanks a lot for your comment!

- For Indexing, it seems to be fine the way I correct it for Eigen https://github.com/openMVG/openMVG/pull/1972/commits/b7457dfc46be92916f637361a108258a4204a78b.  Eigen default is column major.
let V a 3x3 Matrix as
```     
       0 3 6
V =    1 4 7
       2 5 8
```
```
        const double u1 = V(0) - s*V(1);  //=0 - s*1
        const double u2 = V(3) - s*V(4);  //=3 - s*4
        const double u3 = V(6) - s*V(7);  //=6 - s*7
```
so I don't know for CVL but with Eigen as you can see this code "substract" rows, not columns as it should (=> u = V(:, 1) - sV(:,2))

- Yep I know const does not optimize code but I use it for its semantic meaning and think it's harmless (note I'm not using constexpr but only const). 

- For vector I think I got your point, I will try a proposal, but It could breaks some kind of implicit design guidelines we had in the lib. For matrix vs quad, we already discuss about that and it's a design choice, rot matrix are used all over the place in the lib.

midjji:
Huh,
that might actually explain a few issues I had with Eigen in the past !

Assuming you also provide both pnp_ransac it would be easy to provide two
variants of the p3p, with the return vector one calling the vector output
param one.

Also you should probably wrap the p3p in a wrapper which verifies each
solution, as while the algorithm should never return an incorrect solution,
that is only guaranteed when the input is not degenerate. And as a randomly
drawn group of correspondences can be degenerate, it happens. If you then
use if(error>threshold) ++inliers; it will count nan as a inlier.

PR #1971, opened 2021-11-29
Fix build on non-x86

pkubaj:
Tested on FreeBSD/powerpc64, may still have issues with other architectures.

### DIFF
--- a/src/openMVG/matching/metric_simd.hpp
+++ b/src/openMVG/matching/metric_simd.hpp
@@ -17,7 +17,9 @@
 #include <numeric>
 
 #include <cstdint>
+#if defined(__amd64__) || defined(__i386__)
 #include <immintrin.h>
+#endif
 
 namespace openMVG {
 namespace matching {


pmoulon:
Thank you for this pull request.
What would happen if you just use the #ifdef OPENMVG_USE_AVX2 || OPENMVG_USE_AVX at this line?
I'm also curious to find someone test us with a compilation on a ARM platform build

pkubaj:
I tested my patch on armv7, builds fine. ifdefing for AVX should also be fine.

pmoulon:
Could you teach me how do you tested armv7 compilation?

It would be a nice addition if we could have that in the GitHub workflows

pkubaj:
I just built `misc/openmvg` port on my BeagleBone Black.

pmoulon:
@pkubaj Could you test https://github.com/openMVG/openMVG/commit/5f34f562f3932988b91610f433a22963fbb96e76 on your system and let us know if it worked all right?
Thank you

pmoulon:
Fixed by #1979

PR #1978, opened 2021-12-04
[Build] Modernize cmake openmp usage

pmoulon:


rjanvier:
I think it lacks something like ` add_compile_definitions(-DOPENMVG_USE_OPENMP)` when OpenMP is found else the pre processor won't be able to expose source code (where OPENMVG_USE_OPENMP is defined) to the compiler. What do you think?

pmoulon:
You are right. I will fix this!

PR #2053, opened 2022-04-22
Trifocal tensor solve library

gandrademachine:
# Trifocal tensor solve library

The trifocal solve library is a tool for computing camera relative pose solutions from three images. It requires a minimum of three SIFT correspondences seen in all images, rather than five points in two views. This can be used to initialize SfM pipelines for cases where two view initialization fails or if is not robust enough due to features or geometric configuration.
 This is part of a larger ongoing project with supervision of Pierre Moulon (@pmoulon ) and Ricardo Fabbri (@rfabbri ) to introduce trifocal solvers to openMVG and also other solvers that utilize SIFT feature orientation such as P2Pt (pose from only two points) instead of P3P.  

This PR includes in `multiview/trifocal`:

* Solver
* Error function 
* Robust Solve Kernel for RANSAC
* Unit tests 

## Dependencies

* MiNuS - Included as submodule in `dependencies/minus`. This is a Minimal problem Numerical continuation Solver engine from CVPR 2020 [1]

## How to use it 

First use `git submodule update --init --recursive` to populate MiNuS dependency, then check the unit tests. A sample application is already working and it will be proposed in another PR once this initial(and more reliable) PR is incorporated. 

## Speed

At the moment each solve takes about 500 ms therefore it can be used to offline initialize scenes that cannot be otherwise reconstructed at all. 

## Reach us out

* https://github.com/gandrademachine
* https://rfabbri.github.io

## References

Please cite the following work 

[1] Trifocal Relative Pose from Lines at Points and its Efficient Solution,
CVPR 2020 Ricardo Fabbri, Timothy Duff, Hongyi Fan, Margaret Regan, David
de Pinho, Elias Tsigaridas, Charles Wampler, Jonathan Hauenstein, Peter
Giblin, Benjamin Kimia, Anton Leykin and Tomas Pajdla

[2] Robust Structure from Motion using the Trinocular Geometry of SIFT features, Gabriel Andrade, undergraduate final project, Polytechnic Institute at Rio de Janeiro State University IPRJ-UERJ, advisor: Professor Ricardo Fabbri

PR #2061, opened 2022-05-08
2-point pose solver using SIFT orientation

ANK-dev:
P2Pt is a solver for computing camera pose solutions for one image relative to a partial 3D reconstruction, similar to P3P, but using only two SIFT features (exploiting SIFT orientation). This enables more robust SfM, especially in the cases of the scarcity of features that can occur for smooth man-made objects or small overlap.

This is part of a larger ongoing project with supervision of Pierre Moulon (@pmoulon) and Ricardo Fabbri (@rfabbri) to introduce solvers that exploit SIFT feature orientation to openMVG such as P2Pt (pose from only two points) instead of P3P, and also trifocal solvers that can determine relative pose from 3 point correspondences across 3 images rather than 5 of the 5-point algorithm (also using SIFT orientation), see PR #2053.

This PR includes in `multiview/`:

* Solver

Speed
-----

At the moment each solve takes < 70 microseconds per solve on a standard computer.

Reach us out
------------

* https://github.com/ank-dev
* https://rfabbri.github.io

References
----------

Please cite the following work:
Camera Pose Estimation Using Curve Differential Geometry,
IEEE Transactions on Pattern Analysis and Machine Intelligence, 2020,
Ricardo Fabbri, Peter J. Giblin and Benjamin Kimia

### DIFF
--- a/src/openMVG/multiview/solver_resection_p2pt_fabbri.cpp
+++ b/src/openMVG/multiview/solver_resection_p2pt_fabbri.cpp
@@ -1,0 +1,2741 @@
+// This file is part of OpenMVG, an Open Multiple View Geometry C++ library.
+
+// Copyright (c) 2020 Ricardo Fabbri, Ariel Kovaljski
+//
+// Author: Ariel Kovaljski and Ricardo Fabbri
+// Rio de Janeiro State University
+
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+#include "openMVG/multiview/solver_resection_p2pt_fabbri.hpp"
+#include "openMVG/multiview/projection.hpp"
+
+#include <array>
+#include <complex>
+
+namespace openMVG
+{
+namespace euclidean_resection
+{
+  
+// At most 8 solutions with positive depth, TODO: assert if longer
+static constexpr unsigned TS_MAX_LEN = 8;
+static constexpr unsigned RT_MAX_LEN = (TS_MAX_LEN * TS_MAX_LEN);
+
+template <typename T=double>
+class p2pt { // fully static, not to be instantiated - just used for templating
+	public:
+  static void pose_from_point_tangents(
+    const T gama1[3], const T tgt1[3],
+    const T gama2[3], const T tgt2[3],
+    const T Gama1[3], const T Tgt1[3],
+    const T Gama2[3], const T Tgt2[3],
+    T (*output_RT)[RT_MAX_LEN][4][3],
+    unsigned *output_RT_len,
+    T *output_degen
+  );
+};
+
+template<typename T>
+struct pose_poly {
+	T A0, A1, A2, B0, B1, B2, B3, C0, C1, C2, C3, C4,
+		E0, E1, E2, F0, F1, F2, F3, G0, G1, G2, G3, G4,
+		H0, H1, H2, H3, H4, J0, J1, J2, J3, K0, K1, K2, K3,
+		L0, L1, L2, alpha, beta, theta, sth, cth;
+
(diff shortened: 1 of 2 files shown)


pmoulon:
Just launch CI, let's wait and see what it will say ;-)
Thank you @ANK-dev Let me take a look during the week and merge it.

pmoulon:
@ANK-dev CI looks ok on Unix, but it is failing on Windows
`solver_resection_p2pt_fabbri.cpp(178,41): error C2146: syntax error: missing ';' before identifier '__attribute__' `
You can look to a similar fix I made for MINUS here https://github.com/rfabbri/minus/commit/379be504c815852c962a58fa28d4922452d6cb41#diff-04cfd44a7ea0ddf6113b3b09c1e176f832c0b14942972b228736e839761cead4R21

ANK-dev:
Just pushed a fix, let's hope it works.

rfabbri:
Seems the remaining CI error is unrelated to the PR.

pmoulon:
Looks good now ☺️

src/openMVG/multiview/solver_resection_p2pt_fabbri.cpp
```diff
@@ -0,0 +1,2741 @@
...
+			sigmas2[i][sigmas1_len[i]++] = sigma2_p.real();
+		}
+		// if (sigmas1_len[i] == 0) // isempty(sigmas1[i])
+		// std::cerr << "no sigma1, sigma2 pair satisfies the 3rd constraint" << std::endl;
+	}
+}
+
+template<typename T>
+inline void
+invm3x3(const T (&input_m)[3][3], T (&output_m)[3][3])
```
pmoulon, line 2599: Why not use Eigen?

src/openMVG/multiview/solver_resection_p2pt_fabbri.hpp
```diff
@@ -0,0 +1,52 @@
...
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#ifndef OPENMVG_MULTIVIEW_RESECTION_P2Pt_FABBRI_HPP
+#define OPENMVG_MULTIVIEW_RESECTION_P2Pt_FABBRI_HPP
+
+#include "openMVG/multiview/two_view_kernel.hpp"
+#include "openMVG/multiview/solver_resection_metrics.hpp"
+
+#if defined(_MSC_VER)
```
pmoulon, line 17: Should move to CPP since it is not used in the hpp file

src/openMVG/multiview/solver_resection_p2pt_fabbri.cpp
pmoulon: You could easily map memory from your memory pattern to Eigen using map

pmoulon (commented):
Added some minor comments

src/openMVG/multiview/solver_resection_p2pt_fabbri.cpp
rfabbri: The only reason is that using Eigen would add an additional dependency on the original code for p2pt (http://github.com/rfabbri/diffgeom2pose). The way it is, it is slightly easier to copy back changes to/from the original diffgeom2pose repository.
We can use eigen no problem for OpenMVG if you prefer.

src/openMVG/multiview/solver_resection_p2pt_fabbri.hpp
rfabbri: done.

src/openMVG/multiview/solver_resection_p2pt_fabbri.cpp
pmoulon: I'm good with this. Now I have historical context. We can replace later once we have functional test with SFM pipelines.

PR #2065, opened 2022-05-15
Add implementation of linear global translation (LiGT) - Global translations from global rotations and track observations

pmoulon:
## What's new
Implementation of LiGT from [1]. An alternative to [Translation_Structure_L1_ConstraintBuilder](https://github.com/openMVG/openMVG/blob/develop/src/openMVG/linearProgramming/lInfinityCV/tijsAndXis_From_xi_Ri.hpp#L66) that does not compute the 3D coordinates of the points.

This implementation is supported by the Inertial and Visual Fusion (VINF) research group at Shanghai Jiao Tong University https://www.researchgate.net/lab/Inertial-visual-Fusion-VINF-Yuanxin-Wu

## How to use it?
`openMVG_main_SfM --translationAveraging 4`

[1] A Pose-only Solution to Visual Reconstruction and Navigation
Qi Cai, Lilian Zhang, Yuanxin Wu, Wenxian Yu, Dewen Hu

Details:
- Add [Spectra](https://github.com/yixuan/spectra) as a third-party to OpenMVG to solve sparse Eigen Value problems


pmoulon:
Seems like there is an issue with Spectra on mac https://github.com/pose-only-vision/openMVG/runs/6442451077?check_suite_focus=true#step:6:302

pmoulon:
Closing this PR, will re-open when we will have an understanding of the failure on Mac

Issue #1241, opened 2018-02-25
Incremental SfM v2


pmoulon:
Implement a new Incremental SfM pipeline (`SequentialSfMReconstructionEngine2`) with the following features:

- The reconstruction initialization is done thanks to an abstract interface, `SfMSceneInitializer`.
- The triangulation stage considers the entire scene tracks.
- The resection stage is based on 2d-3D matching confidence.
- The reconstruction can start from existing camera poses.

This new engine is easier to read and to customize. This new engine has been designed after reading [1], [2] & [3].

`SequentialSfMReconstructionEngine2` is:
- **fast**:
  - Since it localizes images as soon as it can, fewer Bundle Adjustment steps are observed than in `SequentialSfMReconstructionEngine`.

- **flexible**:
  - The engine can extend a partial reconstruction, you can call this engine on the results of any other SfM Engine. For example, you can run `GlobalSfM`  (to obtain the pose of the camera triplets) and then run `SequentialSfMReconstructionEngine2` to localize the remaining images.
  - You can now initialize the reconstruction with a n-view reconstruction (`Stellar` [2]) and provide a very stable seed for the reconstruction.

For the moment three `SfMSceneInitializer` are implemented:
- `SfMSceneInitializer`
  - Keep the existing poses. -> extend a previous reconstruction.
- `SfMSceneInitializerMaxPair`: 
  - Initialize a 2-view reconstruction (the relative pose with the most of inliers).
- `SfMSceneInitializerStellar`
  - Initialize a stellar reconstruction (a n-view reconstruction with edge connected to a central unique pose: i.e here a 5 pose stellar configuration defined by 4 relative pairs {{0,1}, {0,2}, {0,6}, {0,10}}. see [2])

**Looking forward to the community feedback on this new engine.**

[1] **Batched Incremental Structure-from-Motion.** Hainan Cui, Shuhan Shen, Xiang Gao, Zhanyi Hu. (3DV 2017).
[2] **Global Structure-from-Motion by Similarity Averaging.** Zhaopeng Cui and Ping Tan. (ICCV2015)
[3] **Structure-from-Motion Revisited**, Schonberger, Johannes Lutz and Frahm, Jan-Michael. (CVPR2016).

Here is an example of a stellar reconstruction computed from 27 relative pose pairs and 223 relative scales, the Stellar estimation results in 28 images localized to start the SfM process.
[...]

Issue #1756, opened 2020-06-09
Issues with Pose Priors


csparker247:
I've been trying to import pose priors from an automatic acquisition rig in order to make my results more reliable across different objects, but I've been having issues with the reconstruction workflow. The scanner is a custom rig with five cameras mounted to a motorized XYZ stage. The system is intended to capture flat-ish surfaces (like topographical models) so all cameras point down at the ground plane. The cameras are arranged in a cross pattern where the central camera faces straight down, and the wing cameras point down at a 45 deg angle:
```
Top-Down View
   N
W  C  E
   S 

Side View (X-axis)
 W    C    E  
 \    |    /

----ground----
```

The system runs step-and-shoot acquisition along a customizable path (generally a grid pattern), producing 5 images at discrete acquisition points along the path. The acquisition points are set so that every point in the interior scan area is seen by all five cameras. The final dataset includes the images for all five cameras, the radial (k3) distortion parameters for the cameras, the stage position at each acquisition point, and the relative offset of the cameras from the stage position. I have custom code which converts this metadata into an `sfm_data.json` file. The work I did on #1702 was so that I could verify that my camera poses were being converted correctly. 

I know there's likely to be some error in my camera poses due to both measurement and mechanical errors in the XYZ stage, so I need to do some sort of automatic adjustment on the scene to minimize reprojection error. Following advice that I've seen in other discussions, my workflow has been as follows:

1. Compute features
2. Use frustum filtering and initial poses to create pair list
3. Compute matches with pair list
4. Reconstruct (pick one):
    - (a) ComputeStructureFromKnownPoses + bundle adjustment
    - (b) IncrementalSfM2 with `-f NONE -S EXISTING_POSE -P`

When I run 4a, the filtering step prior to BA removes all of my poses and I get an empty scene. If I disable BA, I get a scene where all of the feature points are dispersed incoherently below the cameras. Not surprising since I assume my camera positions are off:
![sfkp00](https://user-images.githubusercontent.com/1434526/84197255-9f5d3900-aa6f-11ea-8829-08a5969f3362.png)

When I run 4b, my cameras are adjusted way out of position and my feature points do not seem any more coherent than they did with 4a:
![sfm200](https://user-images.githubusercontent.com/1434526/84197587-32966e80-aa70-11ea-99d8-8446063aa66e.png)

So now I'm left not really sure where my problem lies. I see a number of questions that are open:
1. Is this the correct workflow for pose priors in OpenMVG?
[...]

PR #2070, opened 2022-05-19
[Experimental] Add a new SfM pipeline based on stellar reconstruction

pmoulon:
History and background:
- #1241 introduced a method to recover global motions of a `Stellar / n-uplets pod` (one central image + n satellites) as a convenient way to bootstrap an Incremental SfM pipeline from `N images` instead of 2, we are introducing here a SfM pipeline using `n-uplets` as input of a Global SfM pipeline.

# What is this Stellar SfM pipeline?

- We are now introducing a SfM pipeline using n-uplets over the view-graph to a GlobalSfM pipeline.

We see this pipeline as a natural evolution or OpenMVG research progress for SfM due to the following fact:
- We realized that using triplets for GlobalSfM was helping support forward motions and finding much more relative translations to feed for translation averaging.
- We now think that using n-uplet is a natural extension to enforce more consistency and thigh stable results to compare relative motions in epipolar view graph.

**Pros:**
- Using `n-uplets` or `stellar pods` allows us to refine local motions and make relative motions consistency tighter; resulting in improved robust behavior on Rotation and Translation averaging results.
- Adding view graph simplification to this pipeline is extremely easy and allows to leverage often only 50% of the edges to perform very nice reconstructions. We have used here some similar concept as `skeletal graph` but adapted to `n-uplets / stellar` configuration by selection X Maximum Spanning tree (forcing local stars configurations).

**Cons:**
- In an ideal world we would have accurate 2-view rotation and translation at the first stage... but as we know since these steps rely on approximate intrinsics parameters (approximate focal and unknown distortion factors), those relative motions are not perfect and often contaminated by invalid relative motion (due to repetitive patterns or local structure match).

You can find below a short summary comparison of the various reconstruction algorithms:

image

# How to test it?

` $ openMVG_main_SfM -i <sfm_data.json> -m <match_path> -o <output_reconstruction_folder> --sfm_engine STELLAR`

# How can you provide us with feedback?

- Please report on this PR your reconstruction and your observations (timing, quality of the reconstruction) ...
- This is an experimental pipeline, so don't expect it to be more accurate, and faster on all your datasets, but with your feedback, and contribution we are looking forward make it really robust/solid!

Thank you for @rjanvier for his contribution to this development and for helping stabilize the stellar reconstruction algorithms.


csparker247:
Alright, I've had a chance to run comparison recons on the same dataset. As backstory, my scanning setup is essentially the one I described [way back in 2020](https://github.com/openMVG/openMVG/issues/1756#issue-635751608): five cameras which are fixed relative to each other, but which are traveling a grid path on the XY axis above a mostly planar object. The dataset is 780 images (5MP each). My motion is motor controlled, so I have estimates for the extrinsics for each image which haven't ever been good enough to triangulate on their own. Instead, I initialize `sfm_data.json` with those extrinsics as a "hint" and run the standard SfM pipeline. One misconfiguration on my part for this dataset is that the imported extrinsic has an identity rotation and the center is in the correct region of the world coordinate frame, but not where the camera actually was. Historically, global reconstruction has produced better results than either incremental version, so that's what I use exclusively. I have mostly been using OpenMVG v1.6 and just recently updated to v2.0, so I compare global sfm in each of those versions against stellar.

## Reconstruction commands
- Global (v1.6)
     ```shell
    openMVG_main_GlobalSfM -i sfm_data.json -m matches_dir -o recon_dir -M matches_dir/matches.f.bin
    ```
- Global+Robust (v1.6): 
    - In v1.6, I always gotten better results if I followed SfM with a robust triangulation using ComputeStructureFromKnownPoses. I'm still testing whether this is helpful or just a waste of time in v2.0.
    ```shell
    openMVG_main_ComputeStructureFromKnownPoses -i sfm_data.bin -m matches_dir -o sfm_data_structured.bin -f matches_dir/matches.f.bin
    ```
- Global (v2.0)
    ```shell
    openMVG_main_SfM -i sfm_data.json -s GLOBAL -m matches_dir -o recon_dir -M matches_filtered.bin
    ```
- Stellar
    ```shell
    openMVG_main_SfM -i sfm_data.json -s STELLAR -m matches_dir -o recon_dir -M matches_filtered.bin
    ```

## Subjective Results

SfM Methods Comparison


Visually, the reconstructed camera positions are fairly similar. Since I don't have ground truth for this dataset, it's hard to know how close they are to expected, but they seem reasonable relative to the detected scene. A few notes:

- Stellar shows pretty significant "bowing" in the camera positions and scene features.
Stellar bowing
  My gut instinct tells me that this might be caused by differing intrinsic estimations. I have two intrinsics, and the estimates between each method do calculate differing focal lengths, but I'm not experienced to know if they're significantly different in this context. Since Global v2.0 seems visually correct, here are the values from that and Stellar:

|  | Intrinsic 1 (Global v2.0) | Intrinsic 2 (Global v2.0) | Intrinsic 1 (Stellar) | Intrinsic 2 (Stellar) |
| :--- | :---: | :---: | :---: | :---: |
| Focal Length | `4613.07`  | `3717.48`  | `4312.17` | `3440.81` |
| Principal Pt. | `[1235.98, 1001.79]` | `[1221.34, 1006.40]`  | `[1231.27, 1277.60]` | `[1233.89, 1006.36]` |
| K3 | `[-0.23101, 0.24692, -0.38516]`  | `[-0.08339, 0.12214, -0.45791]`  | `[-0.19790, 0.14862, -0.22706]` |  `[-0.08891, 0.10494, -0.29192]` |
 
- Global+Robust has more outlier features than just Global. Stellar has many, many more feature outliers than any other method. Additionally, there are features floating way out in space, behind the cameras? Here's the stellar bounding box if I zoom out:

Stellar outliers 1

Stellar outliers 2

## Empirical Results

To try to put some numbers to this, I've aggregated the results from `SfMReconstruction_Report.html` (or the equivalent) for each reconstruction. `Time (s)` is how long the SfM step took. I don't have this for the v1.6 recons. `Recon scale` is the scale of the reconstruction calculated from known fiducials in the scene after running things through OpenMVS. `Residuals [min|median|max]` are the given operation performed over the corresponding column in the report. For example, Residuals min provides the min of all views in the report's Residuals min column, and Residuals median reports the median of all views in the report's Residuals median column. `[Mean|Median] Residuals *` is similar, except that the mean or median is taken over the given column. Confusing, no?

||Time (s)|Recon Scale|Residuals min|Mean residual min|Median residual min|Residuals median|Mean residual median|Median residual median|Mean residuals mean|Residuals max|Mean residuals max|Median residuals max|
|---|---|---|---|---|---|---|---|---|---|---|---|---|
|Global (v1.6)|-|-|1.76E-07|1.18E-04|6.45E-05|0.2482700|0.2493156|0.2482700|0.3648423|6.8284500|3.8677809|3.7942550|
|Global+Robust (v1.6)|-|0.0523380|3.16E-08|**2.36E-05**|**1.19E-05**|0.2961755|0.2934834|0.2961755|0.4378704|**3.9969700**|**3.7490539**|3.7822950|
|Global (v2.0)|**738**|0.0734842|4.65E-08|1.25E-04|6.68E-05|0.2500920|0.2503197|0.2500920|0.3666414|5.9926200|3.7729333|**3.7325250**|
|Stellar|1050|1.1395956|**1.54E-08**|6.15E-05|2.60E-05|**0.2172065**|**0.2194761**|**0.2172065**|**0.3155448**|153.4100000|6.2293916|4.3827700|

I think the takeaways from this table are:
- Stellar took about 40% longer to compute in this test, but the computation times are not unreasonable for such a large scene.
- Broadly, stellar's residual error seems to be better than or on par with the other methods. One exception is with the maximum residual error. There are some pretty significant feature outliers as shown in the subjective results.
- Stellar produces a recon that's closer to the expected scale than the other methods. I assume this is only because I'm importing some estimate for the camera positions. I vaguely remember seeing a similar thing with IncrementalV2.

pmoulon:
Thank you @csparker247 for this details test and report. Appreciated the level of details!

Just a quick question:
- You mention you have a camera system based on 5 cameras, but you are then listing 2 intrinsics.
A. Are you configuring the same intrinsics for different cameras (based on the notion that they are sharing the same sensor and lens)?
B. Or are you using 2 intrinsics, because you change the focal length between two captures?

Nit: I would just set 1 intrinsic per camera (since each camera can be a little different)

Bowing:
- Using fixed intrinsic calibrations could help solve the problem of bowing (here we see intrinsics and extrinsics compensating for each other) => So you would estimate only the extrinsics, or provide a better guess for the minimization.
  - You could also use the Camera Model `-c 1` to use a lighter distortion model and see if it still fit nicely your camera model
  -  This is also why you see some improvement using global + robust, since robust is using the refined intrinsics to recover better 2-view matches

Floating points:
- Maybe I forgot to add some point filtering methods I'm using in the other pipelines (as showing in the max error)

Scale:
- Stellar SfM does not assume to use your existing poses, so being closer to the real scale is just so luck

I'm glad that you were able to test flawlessly the new pipeline and to notice that the scene seems to have more 3D points.
- Regarding speed (Global vs Stellar) - depending of some graph, stellar or global could be faster, but Stellar will be in general much more robust to outlier data and so fail much less often than the Global pipeline.

csparker247:
Thanks @pmoulon!

>    You mention you have a camera system based on 5 cameras, but you are then listing 2 intrinsics.
>    A. Are you configuring the same intrinsics for different cameras (based on the notion that they are sharing the same sensor and lens)?
>    B. Or are you using 2 intrinsics, because you change the focal length between two captures?
>
>Nit: I would just set 1 intrinsic per camera (since each camera can be a little different)

All five cameras share the same body and 4 have the same type of fixed focal length lens. My assumption was that the intrinsic variations between lenses of the same make and model would not make a significant difference in practice. I have admittedly been very lazy about precalibrating my cameras recently, though, so I suppose I should measure this and make sure my I'm still initializing `sfm_data.json` correctly.

> Bowing:
>
>    Using fixed intrinsic calibrations could help solve the problem of bowing (here we see intrinsics and extrinsics compensating for each other) => So you would estimate only the extrinsics, or provide a better guess for the minimization. 

This tradeoff makes sense. So I should stop being lazy and calibrate my lenses ahead of time 😄. I'll try that and re-run stellar. I will say that I ran a stellar + robust job yesterday and it doesn't appear to improve the bowing. However...

> Floating points:
>
>    Maybe I forgot to add some point filtering methods I'm using in the other pipelines (as showing in the max error)

Adding a robust step did remove the worst of the outlier features, but quite a few still falling behind the imaging plane. Obviously not super important to the discussion here, but thought I should at least mention it:

Stellar + Robust result

> Regarding speed (Global vs Stellar) - depending of some graph, stellar or global could be faster, but Stellar will be in general much more robust to outlier data and so fail much less often than the Global pipeline.

Good to know! I'll keep testing this features as it develops and let you know how it goes!

4CJ7T:
### Command line:
- Stellar
```pwsh
openMVG_main_SfM -i $matchesDir\sfm_data.json -m $matchesDir -o $reconstructionDir -s STELLAR -g 2
```
- Global
```pwsh
openMVG_main_SfM -i $matchesDir\sfm_data.json -m $matchesDir -o $reconstructionDir -s GLOBAL -M $matchesDir\matches.e.bin
```
- IncrementalV2
```pwsh
openMVG_main_SfM -i $matchesDir\sfm_data.json -m $matchesDir -o $reconstructionDir -s INCREMENTALV2
```
- Global+IncrementalV2
```pwsh
openMVG_main_SfM -i $matchesDir\sfm_data.json -m $matchesDir -o $reconstructionDir -s GLOBAL -M $matchesDir\matches.e.bin
openMVG_main_SfM -i $reconstructionDir\sfm_data.bin -m $matchesDir -o $reconstructionDir -s INCREMENTALV2 -S EXISTING_POSE
```
- Stellar+IncrementalV2
```pwsh
```
| (∩^o^)⊃━☆ | Time | Views<br>Reconstructed |
| :---: | :---: | :---: |
| Stellar | 00h 11m 18s | 1037 / 1040 |
| Global | 00h 20m 20s | 1039 / 1040 |
| IncrementalV2 | N/A | N/A |
| Global+IncrementalV2 | 00h 22m 53s | 1040 / 1040 |
| Stellar+IncrementalV2 | N/A | N/A |

Sorry for leaving it half baked. I'll process IncrementalV2 over night and do some more checks tomorrow.

pmoulon:
@4CJ7T Thank you for testing and comparing the different pipeline, this is interesting to see that `Stellar -g 5` is providing a excellent trade off between completeness and timing

Cannot wait to see if you will test on more scenes ;-)

Issue #1271, opened 2018-03-30
image sampler is off by 0.5px


mitjap:
I think that image sample is off by 0.5px.

It all comes to how floating point coordinate system is defined for image coordinates. Integer coordinates are simple - top-left pixel is (`0`,`0`) and bottom-right is (`w-1`,`h-1`) and increase X right and Y down. With floating point coordinates you basically have two options. Center of floating point coordinates system can be center of top left pixel and top-left corner of image is (`-0.5`,`-0.5`) and bottom-right is (`w-0.5`,`h-0.5`). Other options is to have center of coordinate system set to top-left corner of top-left pixel which gives coordinate of upper-left  (`0`,`0`) and bottom-right (`w`, `h`).

Samplers uses first approach where center of floating point coordinate system is in the center of top-left pixel. This by itself is not a problem.

However there is a huge inconsistency with how world points are projected to image. When projecting a point to image image coordinates are in coordinate system with center in top-left corner of top-left pixel.

For example when projecting a point that is in the middle of camera frame and projected to center of image and then sampled, sampler does not return color of centered pixel. See example below.

Another big issue I see is with feature detections. Features are detected in one coordinate system and later when structure points are projected back (as part of SfM for example) their coordinates are compared without consideration of this offset.

```
    int dim = 3;
    int center = dim / 2;

    Image<float> image(dim, dim, true, 0);
    image(center, center) = 1;

    std::cout << "image:" << std::endl << image << std::endl;

    openMVG::cameras::IntrinsicBase *in = new openMVG::cameras::Pinhole_Intrinsic(dim, dim, dim, dim / 2.0, dim / 2.0);
    openMVG::Vec2 x = in->project(openMVG::Vec3(0,0,1));

    openMVG::image::Sampler2d<openMVG::image::SamplerLinear> sampler;

    std::cout << "x:" << x.transpose() << std::endl;
    std::cout << "sampled:" << sampler(image, x.y(), x.x()) << std::endl;
```

```
image:
0 0 0
0 1 0
0 0 0
x:1.5 1.5
sampled:0.25
```
I would expect the sampled value to be `1` (same as center pixel).

mitjap:
These functions would do the transformations:

```
template <class T>
int convertToInt(T x) {
    return static_cast<int>(x);
}

template <class T>
float convertToFloat(T x) {
    return x + 0.5;
}

template <>
float convertToFloat(float x) {
    return x;
}

template <>
float convertToFloat(double x) {
[...]

PR #2087, opened 2022-06-23
Fix inconsistency in spherical Intrinsic_Spherical projection functor

rjanvier:
There is an inconsistency between the way spherical projection is computed in the ceres cost functor and the way it's defined in Camera_Spherical.hpp.
In the ceres functor the coordinates are shifted by -0.5 (top left is (0,0)) and in the Camera_Spherical.hpp file they are not (top left is (-0.5,-0.5)).
Apparently there are some other [inconsistencies](https://github.com/openMVG/openMVG/issues/1271) of this kind in the lib that might need to be addressed.


pmoulon:
Thank you for your contribution. Sorry for the delay to merge.

PR #2092, opened 2022-07-10
(trivial) Remove using namespace std and add std:: qualifications in few places where needed

p12tic:
`using namespace std;` is generally an antipattern as it imports a huge number of common names into the global namespace. We already qualify names with std:: in majority of cases and `using namespace std` is not actually needed most of the time. This commit addresses a small number of remaining cases where std:: is missing and removes all `using namespace std`.

### DIFF
--- a/src/openMVG/exif/exif_IO_test.cpp
+++ b/src/openMVG/exif/exif_IO_test.cpp
@@ -14,7 +14,6 @@
 #include <iostream>
 #include <memory>
 
-using namespace std;
 using namespace openMVG;
 using namespace openMVG::exif;
 
--- a/src/openMVG/features/features_test.cpp
+++ b/src/openMVG/features/features_test.cpp
@@ -18,9 +18,6 @@
 
 using namespace openMVG;
 using namespace openMVG::features;
-using namespace std;
-
-using std::string;
 
 // Define a feature and a container of features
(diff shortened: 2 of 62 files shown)


src/openMVG/exif/sensor_width_database/sensor_width_camera_database.txt
```diff
@@ -15,6 +15,7 @@ Acer CS-5531;5.75
 Acer CS-6530;5.75
 Acer CS-6531;5.75
 Acer CU-6530;5.75
+samsung SM-G973F;5.75
```
pmoulon, line 18: Could we remove this modification from this PR?

pmoulon (commented):
Thank you for your commit.
I made just a minor comment to exclude from this PR the modification of the sensor database.
I also launched the continuous integration and we will see what various compiler says ;-)

src/openMVG/exif/sensor_width_database/sensor_width_camera_database.txt
p12tic: I'm sorry, this was included due to an oversight. Fixed now.

pmoulon:
Thank you. Everything looks good. Thank you for the PR, the contribution. Looking forward your next contribution 😀

PR #2101, opened 2022-08-17
Adding missing GoPro cameras HERO8, HERO9, and HERO10.

alkavan:
* confirmed all same sensor size, and pipeline works.


pmoulon:
Thank you for the contribution

PR #2123, opened 2022-10-26
Colmap ID parity

4CJ7T:
- Starts CAMERA_ID and IMAGE_ID at 1, matching Colmap's export.
- Adds compatibility with existing Colmap scripts counting on ID starting at 1.


pmoulon:
Note, for your information it is using the camera Id and observations Id that are defined by OpenMVG and their JSON files.
The camera Id can be whatever the user wants, by default it starts at 0 to n, but if a users wants to user their own id, like hashes they can to.
So I wonder if the right way would not be to build first a hash_map that map the existing ID to new contiguous one.

Can you show the source of truth specifying that Colmap needs {1-n} mapping?

4CJ7T:
Right, it would mess with existing ambiguous IDs.

PR #2147, opened 2023-01-19
Prefer CMake Config when looking for Flann

### LINKED ISSUE #1265: FindFlann.cmake missing linker flag -llz4.

StefanBruens:
Latest flann may link to e.g. liblz4, and the FindFlann module does not handle transitive lib dependencies.

In case external flann is required try the upstream CMake Config first, and fall back to the legacy lookup.

Fixes #1265.


pmoulon:
Thank you for your PR @StefanBruens.
I kick off the CI, so we will see if everything is looking good on compilation and unit test (but not sure I installed FLANN in https://github.com/openMVG/openMVG/blob/develop/.github/workflows/compile_and_run_test_external_deps.yml)
Nit: We are using FLANN in header only mode

pmoulon:
@StefanBruens Thx you for your contribution. Merged!

PR #2148, opened 2023-01-21
Fix library install directories

StefanBruens:


pmoulon:
Thx you for your contribution.
- Can you tell what is the behavior of GNUInstallDirs for windows?
- - Why stlplus should always be build in static?

StefanBruens:
For Windows, `CMAKE_INSTALL_LIBDIR` is `lib`, i.e. this should be unchanged.

pmoulon:
Thx you for confirming!

StefanBruens:
Otherwise, the unversioned and more or less openMVG specific `stlplus` library must be installed into a system directory (or one has to fiddle with RUNPATH etc).
It also mimics how the other vendored libraries are build.

IMHO longterm this stlplus should be replace with std::filesystem (C++17).

src/CMakeLists.txt
```diff
@@ -595,11 +597,10 @@ message("\n")
 # INSTALL RULES
 # ==============================================================================
 
-install(EXPORT openMVG-targets DESTINATION lib)
 install(EXPORT openMVG-targets
   FILE OpenMVGTargets.cmake
   NAMESPACE OpenMVG::
-  DESTINATION share/openMVG/cmake
+  DESTINATION ${CMAKE_INSTALL_LIBDIR}/openMVG/cmake
```
pmoulon, line 603: I would still keep the share here (see cmake doc https://cmake.org/cmake/help/book/mastering-cmake/chapter/Install.html)

src/CMakeLists.txt
```diff
@@ -616,7 +617,7 @@ configure_file("${CMAKE_SOURCE_DIR}/cmakeFindModules/OpenMVGConfig.cmake.in"
                "${CMAKE_CURRENT_BINARY_DIR}/OpenMVGConfig.cmake" @ONLY)
 
 install(FILES "${CMAKE_CURRENT_BINARY_DIR}/OpenMVGConfig.cmake"
-        DESTINATION share/openMVG/cmake)
+        DESTINATION ${CMAKE_INSTALL_LIBDIR}/openMVG/cmake)
```
pmoulon, line 620: See above, I would keep the share

src/openMVG/cameras/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ set_property(TARGET openMVG_camera APPEND PROPERTY
 
 target_compile_features(openMVG_camera INTERFACE ${CXX11_FEATURES})
 target_link_libraries(openMVG_camera INTERFACE openMVG_numeric ${cereal_TARGET} ${OPENMVG_LIBRARY_DEPENDENCIES})
-install(TARGETS openMVG_camera DESTINATION lib EXPORT openMVG-targets)
+install(TARGETS openMVG_camera EXPORT openMVG-targets)
```
pmoulon, line 8: Just to confirm, you see no need of lib since it is a headeronly library?

src/openMVG/system/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ target_include_directories(openMVG_system PUBLIC $<BUILD_INTERFACE:${CMAKE_SOURC
 target_compile_features(openMVG_system INTERFACE ${CXX11_FEATURES})
 set_target_properties(openMVG_system PROPERTIES SOVERSION ${OPENMVG_VERSION_MAJOR} VERSION "${OPENMVG_VERSION_MAJOR}.${OPENMVG_VERSION_MINOR}")
 set_property(TARGET openMVG_system PROPERTY FOLDER OpenMVG/OpenMVG)
-install(TARGETS openMVG_system DESTINATION lib/ EXPORT openMVG-targets)
+install(TARGETS openMVG_system DESTINATION ${CMAKE_INSTALL_LIBDIR}/ EXPORT openMVG-targets)
```
pmoulon, line 9: for consistency with the other, remove the `/` at the end

src/openMVG/cameras/CMakeLists.txt
StefanBruens: Correct.

src/CMakeLists.txt
pmoulon: So lib/share would be an appropriate path from your knowledge. We can check the behavior on a local folder install and see if consistent with the doc https://github.com/openMVG/openMVG/blob/develop/BUILD.md#using-openmvg-as-a-third-party-library-dependency-with-cmake

StefanBruens: If you meant `lib` _or_ `share`, then yes, but not `lib/share` (unless `lib/` is actually part of the INSTALL_PREFIX).

(`|` in the link above is part of a regex and means "or").

But `share` is only allowed for architecture independent packages (e.g. header only libraries).

StefanBruens: Convinced this is correct and does not cause any breakage?

StefanBruens: In the documentation, `<userprefix>/share/cmake` should become `<userprefix>/lib/cmake`.

(In case `CMAKE_INSTALL_PREFIX` is not one of the standard system prefixes, e.g. `/usr/` or `/usr/local/`, CMake defaults to just `lib`.)

pmoulon:
@StefanBruens Thx you for your contribution. Merged!

PR #2167, opened 2023-02-19
[software/import] Add support of Multiface dataset

kenshi84:
Add the ability to load a Multiface dataset [1] directly as an openMVG sfm_data file.

[1] Multiface: A Dataset for Neural Face Rendering. Cheng-hsin Wuu et al. arXiv 2022.
https://github.com/facebookresearch/multiface

### DIFF
--- a/src/software/SfM/import/SfM_StructurePipeline.py.in
+++ b/src/software/SfM/import/SfM_StructurePipeline.py.in
@@ -30,7 +30,7 @@ import sys
 
 if len(sys.argv) < 5:
     print ("Usage %s image_dir gt_dir gt_type output_dir" % sys.argv[0])
-    print ("gt_type: %s 1:Strecha's Dataset 2:MiddleBury's Dataset 3:DTU MVS Dataset 4:ETH 3D Dataset 5:Kitti Odometry Dataset" % sys.argv[0])
+    print ("gt_type: %s 1:Strecha's Dataset 2:MiddleBury's Dataset 3:DTU MVS Dataset 4:ETH 3D Dataset 5:Kitti Odometry Dataset 6:BlendedMVS Dataset 7:Multiface Dataset" % sys.argv[0])
     sys.exit(1)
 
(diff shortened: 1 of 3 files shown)


pmoulon:
Thank you for your contribution. Could you share the command line you ran for import and eventually a screenshot showing what you did with a successful import?

kenshi84:
Sure, the dataset content is organized in a way described here:
https://github.com/facebookresearch/multiface/blob/main/documentation/DATA_STRUCTURE.md

Suppose I wanted to run SfM on the first frame of the `E001_Neutral_Eyes_Open` sequence in the entity `002757580`.
Because the dataset stores image sequences in separate directories corresponding to different views, first I needed to re-organize image files so that images at a certain time frame from different viewpoints are stored in the same directory. After cd'ing to the entity root directory `m--20171024--0000--002757580--GHS`, the following Python script does that:

```py
import os
import shutil

for x in os.walk('images/E001_Neutral_Eyes_Open'):
    for name in x[1]:
        shutil.copy('images/E001_Neutral_Eyes_Open/' + name + '/000213.png', 'images/')
    break
```

Then, I can run OpenMVG pipeline as follows:
```sh
python3 SfM_StructurePipeline.py \
path/to/m--20171024--0000--002757580--GHS/images \
path/to/m--20171024--0000--002757580--GHS/ \
7 \
path/to/outdir
```

Here's a screen shot of the resulting colored point cloud:
Screen Shot 2023-02-20 at 11 48 13

pmoulon:
Great. Thx you. To deal with multiple folder I often use symlinks but copy is working great too. I will try repro and merge soon!

pmoulon:
I was able to repro.
It would be worth documenting how to export a given `pose` via your script, or automating it with detecting all pose ID and export them to various folders with symlinks.

For the fun, I tried OpenMVS on the scene imported in openMVG

Thx you @kenshi84 for this nice contribution!

Point cloud:
image

Mesh:
image

Issue #2198, opened 2023-05-16
ComputeFeatures doesn't utilize many CPU cores fully


bouk:
Hi, I'm running openmvg in a pipeline on an EC2 c6a.48xlarge machine which has 192 cores. When looking at `htop` I can see that the cores spike up to full usage and then fall back to only one core being at 100% constantly, for a total load avg of only ~30% (htop screenshot below). I'm speculating that this is caused by [the `dynamic` openmp scheduling
](https://github.com/openMVG/openMVG/blob/b9024c3117f4df283bb58c6474ee2c224e6c884e/src/software/SfM/main_ComputeFeatures.cpp#L266) having a lot of overhead.

I'll do some testing using a different scheduling strategy but I just wanted to open this issue to track it

```
INFO: [main_ComputeFeatures.cpp:120]  You called :
/run/current-system/sw/bin/openMVG_main_ComputeFeatures
--input_file /root/output/sfm/matches/sfm_data.json
--outdir /root/output/sfm/matches
--describerMethod SIFT
--upright 0
--describerPreset HIGH
--force 0
--numThreads 0

INFO: [loggerprogress.hpp:79] [- EXTRACT FEATURES -] 10%
INFO: [loggerprogress.hpp:79] [- EXTRACT FEATURES -] 20%
INFO: [loggerprogress.hpp:79] [- EXTRACT FEATURES -] 30%
INFO: [loggerprogress.hpp:79] [- EXTRACT FEATURES -] 40%
INFO: [loggerprogress.hpp:79] [- EXTRACT FEATURES -] 50%
INFO: [loggerprogress.hpp:79] [- EXTRACT FEATURES -] 60%
INFO: [loggerprogress.hpp:79] [- EXTRACT FEATURES -] 70%
INFO: [loggerprogress.hpp:79] [- EXTRACT FEATURES -] 80%
INFO: [loggerprogress.hpp:79] [- EXTRACT FEATURES -] 90%
INFO: [loggerprogress.hpp:79] [- EXTRACT FEATURES -] 100%
INFO: [main_ComputeFeatures.cpp:343] Task done in (s): 707
```

Htop constantly cycles between these three screens:

<img width="3008" alt="Screenshot 2023-05-16 at 12 47 19" src="https://github.com/openMVG/openMVG/assets/97820/aed1223b-751a-4cf0-9863-2c9910160b9f">
[...]

Issue #1730, opened 2020-05-01
 openMVG_main_ComputeFeatures multi threading 


YouYue123:
Hi,

I am running OpenMVG pipeline with around 350 images in a 16-core machine. openMVG_main_ComputeFeatures setting is as below
```
--describerMethod AKAZE_FLOAT
--upright 0
--describerPreset HIGH
--force 0
--numThreads 0
```
My understanding is --numThreads 0 is for max_thread right?

But I observed the CPU usage is not always utilizing the full 16 cores. Most of the time it is only using 1 CPU as below.

![Screenshot 2020-05-01 at 9 58 07 AM](https://user-images.githubusercontent.com/8738486/80791634-8f118e80-8b92-11ea-9dc9-ed0c43fed258.png)

May I know is this by design or it is related to the feature I choose?  Shall I explicitly input the numThread?

Thanks

rperrot:
Did you compile with openMP activated ? Some of the parts are mono-thread, some are multithreads. AKAZE computation is mainly mono-thread (due to the nature of the diffusion process). 

The --numThread corresponds to multiple computation at once (ie: multiple images are described in parallel). However if you don't have openMP enabled this is not used. (That's the case on macOS with apple compiler).

rperrot:
Note: if you use --numThreads = 0 this forces using the maximum of threads available.

YouYue123:
Thanks for the explanation!

YouYue123:
Thanks for the reply and explanation.

Yes. I compiled the source with openMP activated. Thanks for explaining --numThread

PR #2216, opened 2023-07-05
Enable multithreading when numThreads == 0.

yanxke:
It seems like multiple people have encountered this bug where the CPU is only 100% on the compute features stage. 
 https://github.com/openMVG/openMVG/issues/2198  https://github.com/openMVG/openMVG/issues/1730

After applying the change from https://github.com/openMVG/openMVG/issues/1730#issuecomment-678771888 it worked for me as well.

### DIFF
--- a/src/software/SfM/main_ComputeFeatures.cpp
+++ b/src/software/SfM/main_ComputeFeatures.cpp
@@ -261,7 +261,7 @@ int main(int argc, char **argv)
         omp_set_num_threads(iNumThreads);
     } else {
         iNumThreads = nb_max_thread - 1;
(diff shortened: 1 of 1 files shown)


src/software/SfM/main_ComputeFeatures.cpp
```diff
@@ -260,7 +260,8 @@ int main(int argc, char **argv)
     if (iNumThreads > 0) {
```
pmoulon, line 260: Would not be better to have the value `iNumThreads` updated once with a IF, and then call once `omp_set_num_threads(iNumThreads);`

PR #2217, opened 2023-07-05
Modernize Python pipeline scripts and add optional parameters.

yanxke:


pmoulon:
Since it changes how the binary are called we would also have to update the documentation
See here https://github.com/openMVG/openMVG/blob/develop/docs/sphinx/rst/software/SfM/SfM.rst#openmvg-sfm-pipelines-demo

yanxke:
> 

Sure, added.  The default behavior is the same as before as the new parameters are optional.

PR #2257, opened 2023-10-24
Update Dockerfile

awarebayes:
Python-minimal is no longer in ubuntu repos


pmoulon:
Thank you for the update @awarebayes

PR #2259, opened 2023-10-27
fix: compilation for ubuntu22.04 arm64

jujimeizuo:
error: On machines not built on x86-64 architecture, the `cpuid.h` file does not exist

```bash
/home/jujimeizuo/Workspace/openMVG/src/openMVG/system/cpu_instruction_set.hpp:18:12: fatal error: cpuid.h: No such file or directory
   18 |   #include <cpuid.h>
      |            ^~~~~~~~~
compilation terminated.
```

fixed: https://github.com/openMVG/openMVG/issues/1881#issuecomment-886149844, `#if defined __aarch64__` check architecture

### DIFF
--- a/src/openMVG/system/cpu_instruction_set.hpp
+++ b/src/openMVG/system/cpu_instruction_set.hpp
@@ -15,7 +15,9 @@
 #if defined _MSC_VER
   #include <intrin.h>
 #else
+#if not defined __aarch64__
   #include <cpuid.h>
+#endif
 #endif
 
 namespace openMVG
@@ -117,6 +119,9 @@ class CpuInstructionSet
 private:
   static bool internal_cpuid(int32_t out[4], int32_t x)
   {
+    #if defined  __aarch64__
+    return false;
+    #endif
     #if defined __GNUC__
     __cpuid_count(x, 0, out[0], out[1], out[2], out[3]);
     return true;


src/openMVG/system/cpu_instruction_set.hpp
```diff
@@ -117,9 +119,14 @@ class CpuInstructionSet
 private:
   static bool internal_cpuid(int32_t out[4], int32_t x)
   {
+    #if defined  __aarch64__
+    return false;
+    #endif
     #if defined __GNUC__
-    __cpuid_count(x, 0, out[0], out[1], out[2], out[3]);
-    return true;
+      #if not defined __aarch64__
```
pmoulon, line 126: This one should not be needed, since you already return false for aarch64, right?

jujimeizuo: yeah, but the compiler thinks this is an error:

image

pmoulon: So we could have a unique if defined aarch64 return false here and remove the one on top, right?

pmoulon: @jujimeizuo could you update the code to have only one entry with `__aarch64__` so it would ease later maintenance?

jujimeizuo: yep

pmoulon:
Thank you for your contribution

Issue #2267, opened 2023-11-30
A little question about track.hpp


StraySparks:
Line 135-136 in track.hpp, I wonder if "const auto & feat = map_node_to_index[k]; const uint32_t & track_id = uf_tree.Find(feat.second);" make it more reasonable? 
Because in line 188, the input of "uf_tree.Union()" is "map_node_to_index.second", the input of "uf_tree.Find()" should be the same, but "k" (line 135) is the index of "m_vec" in "class flat_pair_map". The code works well in line 135 because the value of "k" and "map_node_to_index.second" is the same when elements are already sorted after allFeatures.emplace() (type "set", line 74, 85), which means "map_node_to_index.sort()" actually did not work in line 100. If "map_node_to_index.sort()" worked, then the value of "k" and "map_node_to_index.second" will  be different.
I wonder if I've figured it out. Looking forward to your answers and thank you!

Best wishes!

pmoulon:
Can you propose the change in a PR so we can investigate and clearly see the change you are proposing?

StraySparks:
> Can you propose the change in a PR so we can investigate and clearly see the change you are proposing?
I will try it later.

pmoulon:
For this last one.. fell free to make the change locally and run the unit test. It seems indeed a bit strange.

PR #2268, opened 2023-11-30
Update tracks.hpp

StraySparks:
I think "k" and "feat.second" have the same value in line 135-136 of  "tracks.hpp", but they represent different meanings. Reference to the input of "uf_tree.Union()" in line 118, "feat.second" might be the more logical choice for the input of "uf_tree.Find()".
The reason why "k" and "feat.second" have the same value but different meanings can be find in the issue: #2267


pmoulon:
All automated check passed ok. Let me double look to this and test it on various scene and come back to you early next week.

PR #2284, opened 2024-01-12
adding openMVG sample to visualize sfm_data using rerun sdk

khshmt:


pmoulon:
Thank you for your PR. I have also some work in progress that I did not pushed yet to log the growing sfm_data file as it is updated during the SFM process. I will merge both work. 

How is your experience logging images. I got issue on dataset with hundred of images for out of memory. 

Also it could be nice if you were willing to share some reconstruction you made and there rerun preview 😜

khshmt:
Hi Pierre thanks for your response,
I tried the example on two datasets one of them is yours "Sceaux castle"  and the other dataset is for underwater coral reefs but both of them is about 10 images, so I faced no problem because of memory but I will explore bigger dataset soon.

this is a rerun preview for a reconstruction I made 

Screenshot from 2024-01-13 12-35-47

khshmt:
Hi @pmoulon I hope you are doing well,  
I updated merge with a second commit, I hope you try it on the dataset with hundreds of images you mentioned before Which caused out of memory issue

src/openMVG_Samples/visualize_sfm_data/visualize_sfm_data.cpp
```diff
@@ -0,0 +1,126 @@
...
+    }
+  }
+
+  const auto& landmarks = sfm_data.GetLandmarks();
+  std::vector<rerun::components::Position3D> points3d;
+  std::vector<rerun::components::KeypointId> track_ids;
+  std::unordered_map<uint32_t, std::vector<rerun::components::Position2D>>
+      points2d_per_img;
+  for (const auto& landmark : landmarks) {
+    points3d.push_back(rerun::components::Position3D(
```
pmoulon: Minor you can use emplace_back
```
points3d.emplace_back(
        landmark.second.X(0), landmark.second.X(1), landmark.second.X(2));
```

src/openMVG_Samples/visualize_sfm_data/visualize_sfm_data.cpp
```diff
@@ -0,0 +1,126 @@
...
+      const rerun::datatypes::Vec2D resolution{
+          static_cast<float>(view->ui_width),
+          static_cast<float>(view->ui_height)};
+
+      rec.log(camera_entity + view_file_name,
+              rerun::archetypes::Pinhole::from_focal_length_and_resolution(
+                  sfm_data.GetIntrinsics().at(id_intrinsic)->getParams()[0],
+                  resolution));
+      openMVG::image::Image<openMVG::image::RGBColor> img;
+      auto is_img_loaded =
```
pmoulon, line 89: Right way to access the image path is to use the root path in SfM_data
```
      const std::string sImageName = stlplus::create_filespec(sfm_data.s_root_path, view->s_Img_path);
      auto is_img_loaded =
          openMVG::image::ReadImage(sImageName.c_str(), &img);
```

pmoulon (changes requested):
LGTM, just some minor improvement that we can do:
- use emplace_back to be less verbose
- better use `sfm_data.s_root_path` to log the path to the image

pmoulon:
Could be a limitation on my machine
image

Else loading without image is working:
image
image

src/openMVG_Samples/visualize_sfm_data/visualize_sfm_data.cpp
khshmt: I will work on these changes today.
Thanks.

khshmt:
I ran the example using about 140 images and worked well with the images, but I will double check.

src/openMVG_Samples/visualize_sfm_data/visualize_sfm_data.cpp
khshmt: @pmoulon requested changes finished.

pmoulon:
Waiting for CI and I will merge.
Thank you @khshmt for the contribution.
I will also push soon the integration I made to log as the SFM process is working, so you can have a real time preview of where the SFM process is

PR #2294, opened 2024-02-13
Allow control points to be optimized

jvanvugt:
Adds an option to the `Control_Point_Parameter` struct to enable optimization of control points in BA, defaults to false so behavior is unchanged.


pmoulon:
@jvanvugt thank you for the pull request. 
Do you have any examples to share to show how much your control points changed during the process?

jvanvugt:
In my use case, I'm detecting fiducial markers in images and adding them as control points after doing a regular reconstruction. They are triangulated with `TriangulateNViewAlgebraic` and then I run BA to minimize the geometric error. I'm not sure how much the points still move but it would be the difference between the algebraic and geometric error.

PR #2334, opened 2024-06-18
[C++] Remove a redundant copying in saveDescsToBinFile

gongminmin:
* Change the copy to reference.
* Rename the variable from iter to desc, since it's not iterator and the name can be confusing.
* Use const to pass the vectors to saveFeatsToFile and saveDescsToBinFile.

### DIFF
--- a/src/openMVG/features/descriptor.hpp
+++ b/src/openMVG/features/descriptor.hpp
@@ -206,7 +206,7 @@ inline bool loadDescsFromBinFile(
 template<typename DescriptorsT >
 inline bool saveDescsToBinFile(
   const std::string & sfileNameDescs,
-  DescriptorsT & vec_desc)
+  const DescriptorsT & vec_desc)
 {
   using VALUE = typename DescriptorsT::value_type;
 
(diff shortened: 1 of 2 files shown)


pmoulon:
Thank you for the contribution and helping improve our codebase!
Lets ensure the CI is happy and then I will merge.

gongminmin:
> Thank you for the contribution and helping improve our codebase! Lets ensure the CI is happy and then I will merge.

Seems some CI is failing, but not because of my change? The macOS one has
#  error "mathop_sse2.c must be compiled with SSE2 intrinsics enabled"
The Windows one builds but unit tests has 
CMake Error: '-j$' given invalid value '-j$'

gongminmin:
I may know why the CI fails. The macOS-latest is switched to arm so USE_SSE2 may need to be disabled. And in unit test part, the $(nproc) is not defined on Windows. I think just a -j can do for all platforms since it means use all processors.

gongminmin:
I just add a MR https://github.com/openMVG/openMVG/pull/2337 to see if my theory works.

pmoulon:
LGTM, fixing CI and I will merge

PR #2335, opened 2024-06-18
[Exif] Code improvements to Exif_IO_EasyExif

gongminmin:
* Add a destructor to Exif_IO_EasyExif so it can be used as a stack-based object.
* Move bHaveExifInfo_ into EXIFInfoImpl.
* Simplify (*pimpl_).get() to pimpl_->get().

I don't actually know if there is any special reason to use it in unique_ptr<> way, or in (*pimpl_) form. Just feel it can be simplified.


pmoulon:
LGTM, thank you for your contribution

PR #2337, opened 2024-06-19
WIP: [CI] Try to fix CI failures on macOS and Windows

gongminmin:
* Assign bash as the shell to fix the missing $(nproc) in unit tests.
* Turn of USE_* when its related compiler flag is not available.


gongminmin:
My theory of why it fails in macOS-latest is, the CI environment has been changed to ARM only.

1. It doesn't accept match=core2 anymore:
  -- Performing Test check_c_compiler_flag__march_core2
-- Performing Test check_c_compiler_flag__march_core2 - Failed

2. Due to the -DTARGET_ARCHITECTURE=core, sse2 is still added to the list:
list(APPEND _available_vector_units_list "sse" "sse2" "sse3")

3. Since it's in the list, USE_SSE2 is set.
_enable_or_disable(SSE2 "sse2" "Use SSE2. If SSE2 instructions are not enabled the SSE implementation will be disabled." false)

4. Later on when detecting sse2 fails, but it's too late, USE_SSE2 has been set.
AddCompilerFlag("-march=${_flag}" CXX_RESULT _good CXX_FLAGS Vc_ARCHITECTURE_FLAGS)
-- Performing Test check_c_compiler_flag__msse2
-- Performing Test check_c_compiler_flag__msse2 - Failed

So my updated fix for it is add a few lines to turn off USE_* when AddCompilerFlag can't find the flag. Unfortunately I don't have a mac to validate it locally.

pmoulon:
Thank you for helping on this.

pmoulon:
I see the failure on Max as apache arrow not being able to be retrieved via git ..

gongminmin:
Right. Now the Windows one is fixed. On mac, now there is a 
Disabling USE_SSE2 because -msse2 compiler flag is missing
So SSE2 error is avoided on mac. But I don't know what's that arrow error. A glitch on network?

pmoulon:
I will try to have a look during the week end about this

PR #2352, opened 2024-08-26
Fix compilation failure caused by API changes

zhaozigu:
In a Windows environment, after generating the project using CMake, one module fails to compile due to API changes.


pmoulon:
Thank you for the fix, I also updated to use a fixed version of the API to avoid any future break.
I also fix the image size that were not matching (maybe you were having square images for your test cases)

image

pmoulon:
Thank you @zhaozigu  sorry that it took me some time for merging...

Issue #2360, opened 2024-10-04
undefined symbol in svg_features.hpp


Nettal:
function `bool Features2SVG
(
  const std::string & image_path,
  const std::pair<size_t,size_t> & image_size,
  const features::SIOPointFeatures & features,
  const std::string & svg_filename,
  const double feature_circle_radius = 4.0,
  const double stroke_size = 2.0
)`
declared in [svg_features.hpp](https://github.com/openMVG/openMVG/blob/e598c1a1e3fa2d2110587a41d64b72d9049ec4fb/src/openMVG/features/svg_features.hpp#L34C1-L42C3)
but not implement in svg_features.cpp, 
but i find a similar function in [svg_features.cpp](https://github.com/openMVG/openMVG/blob/develop/src/openMVG/features/svg_features.cpp#L51).
possible change the header?

pmoulon:
Hello @Nettal thank you for reporting, feel free to make a PR to fix the issue

PR #2361, opened 2024-10-05
[features]fix undefined symbol in svg_features.hpp

Nettal:
issue #2360 
Features2SVG: remove useless parameter `feature_circle_radius`, update the comment

### DIFF
--- a/src/openMVG/features/svg_features.hpp
+++ b/src/openMVG/features/svg_features.hpp
@@ -49,8 +49,6 @@ bool Features2SVG
  * @param[in] image_size Image size {width,height}.
  * @param[in] features image features.
  * @param[in] svg_filename Path to the output SVG file.
- * @param[in] feature_circle_radius Radius of the circle used to represent
- * the features.
  * @param[in] stroke_size Stroke size used to display the line between the
  * corresponding features.
  */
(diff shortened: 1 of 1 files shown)


fallenmi (approved):
The SIOPointFeatures declaration now matches the overload that is actually defined in `svg_features.cpp`. I reproduced the original failure with a two-translation-unit compile/link oracle against exact merge-base `e598c1a1e3fa2d2110587a41d64b72d9049ec4fb`: the caller compiled from the old header requests the six-parameter symbol, while the implementation exports the five-parameter symbol, so linking fails. The same oracle against exact head `e38c984aee9f55a8cd5330895369603b42d4da70` and current GitHub merge `ab9eef5b286a2a15d682e6027aea774c208b7628` links and runs successfully. The current GitHub merge tree also matches an independent merge into live `develop`.

No blocking findings from me. This upstream head currently has no GitHub check runs, commit status contexts, check suites, or Actions workflow runs, so my verification was local and focused.

Disclosure: I used OpenAI Codex to inspect the exact diff and applicable repository policy, verify the live refs, interactions, workflow layers, implementation signature, and current merge tree, build and run the exact-revision compile/link oracle, and draft this review. I independently checked the evidence and agree with the conclusion.

PR #2364, opened 2024-10-11
[robust_estimation] Add openmp parallelism in other guided matching

Nettal:
Openmp parallel was introduced for GuidedMatching since 2021, but other overloads have not.

### DIFF
--- a/src/openMVG/robust_estimation/guided_matching.hpp
+++ b/src/openMVG/robust_estimation/guided_matching.hpp
@@ -41,7 +41,7 @@ void GuidedMatching(
   //  the smallest distance (smaller than the provided Threshold)
 
   #pragma omp parallel for
-  for (int i = 0; i < xLeft.cols(); ++i) {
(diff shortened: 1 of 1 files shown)


src/openMVG/robust_estimation/guided_matching.hpp
```diff
@@ -41,7 +41,7 @@ void GuidedMatching(
   //  the smallest distance (smaller than the provided Threshold)
 
   #pragma omp parallel for
-  for (int i = 0; i < xLeft.cols(); ++i) {
+  for (size_t i = 0; i < xLeft.cols(); ++i) {
```
pmoulon: I believe int is required here for visual studio compiler

Nettal: Okay, I am not familar with visual studio compiler. and i will change them to int.

pmoulon:
Merged. Thank you for your contribution @Nettal

PR #2392, opened 2025-02-03
A new binary to merge SfM scenes based on the name of the images in each scene 

eokeeffe:
This is pretty useful for larger reconstructions.

Screenshot from 2025-02-03 12-02-38
Screenshot from 2025-02-03 12-02-11
Screenshot from 2025-02-03 12-01-50
Screenshot from 2025-02-03 12-01-07

It takes two SfM scenes and produces an SRT of the positions of the images that are in both scenes. The poses from the second scene are appended to the first scene. SRT is applied to any new poses added to the first scene. SRT is also applied to landmarks. 

I took a nice component from @sebastienchappuis on the camera intrinsic selection, use the intrinsic that gives the best RMSE for the scene. 

At the moment I only assume one camera scene merges, I'll update to have it support multiple camera merges in the future. 

The merge can be done with a final BA, this could probably use a bit more thought, I just copied the parameters for incrementalV1 and global for the moment. You can use without BA for camera setting 1 , the SRT tends to fail for camera setting 1 for in Eigen::umeyama.

Any issues or updates anyone can think I'm happy to have a look at.


eokeeffe:
System works fairly well, simple and does the job

.vscode/settings.json
```diff
@@ -0,0 +1,65 @@
+{
```
pmoulon: Not sure we should take this file in part of the commit

src/openMVG/cameras/Camera_Pinhole.hpp
```diff
@@ -48,15 +48,17 @@ class Pinhole_Intrinsic : public IntrinsicBase
...
     * @param w Width of the image plane
     * @param h Height of the image plane
+    * @param ccdw width of the ccd sensor
     * @param focal_length_pix Focal length (in pixel) of the camera
     * @param ppx Principal point on x-axis
     * @param ppy Principal point on y-axis
     */
     Pinhole_Intrinsic(
       unsigned int w = 0, unsigned int h = 0,
+      double ccdw = 0.0,
```
pmoulon: This introduce an API change, is it necessary?

pmoulon: Maybe disentangle your commit linked to mergeSfM and ReportGSD

pmoulon (changes requested):
Please disentangle your commit linked to mergeSfM and ReportGSD

src/openMVG/cameras/Camera_Pinhole.hpp
eokeeffe: ah, apologies, I'll remove that. 
It was to make it a bit easier to calculate the GSD without having to search the database for the ccdw again

.vscode/settings.json
eokeeffe: It's gone now

src/openMVG/cameras/Camera_Pinhole.hpp
eokeeffe: The changes from ReportGSD have been removed and the branch is up to date with latest develop branch

eokeeffe:
Cleaned up the git history, GSD is fully removed.

eokeeffe:
I've made some improvements to the merging code

- now accepts multiple cameras
- faster merging by fixing the way I was merging the landmarks
- better quality by fixing the BA at the end of a merge

eokeeffe:
This is going to be very useful as I've just finished another module for molecular clustering of the images using combinations of GPS and visual features. This will be very useful for breaking up large datasets into overlapping clusters, then this merging code can be used to rejoin them.

PR #2394, opened 2025-02-08
CMake: Fix OpenMP being expected when not found

### LINKED ISSUE #2393: Installed OpenMVG libraries expect OpenMP when not actually found
I was trying to use the installed libraries in my own CMake project, but hit an issue where `find_package(OpenMVG REQUIRED)` was failing because OpenMP couldn't be found. I thought this was curious, because my OpenMVG install output explicitly stated it couldn't find OpenMP:

```
-- Could NOT find OpenMP_C (missing: OpenMP_C_FLAGS OpenMP_C_LIB_NAMES) 
-- Could NOT find OpenMP_CXX (missing: OpenMP_CXX_FLAGS OpenMP_CXX_LIB_NAMES) 
-- Could NOT find OpenMP (missing: OpenMP_C_FOUND OpenMP_CXX_FOUND) 
```

Looking closer at the CMake output, though, I noticed it later reported that it was still expecting for OpenMP to be enabled:
```
** OpenMVG version: 2.1.0
** Build Shared libs: OFF
** Build OpenMVG tests: OFF
** Build OpenMVG softwares: ON
** Build OpenMVG GUI softwares: OFF
** Build OpenMVG documentation: OFF
** Build OpenMVG samples applications: OFF
** Build OpenMVG openGL examples: OFF
** Enable code coverage generation: OFF
** Enable OpenMP parallelization: ON
** Build OpenCV+OpenMVG samples programs: OFF
** Use OpenCV SIFT features: OFF
** Use LiGT for global translation estimation: ON
```

I suspect that the issue is that this block in OpenMVG's `CMakeLists.txt` isn't actually updating the `OpenMVG_USE_OPENMP` cache variable:
https://github.com/openMVG/openMVG/blob/4e78ed73a575beae88199fc440183a75e03068c5/src/CMakeLists.txt#L182-L187

And subsequently the exported CMake module is incorrectly still looking for OpenMP here:
https://github.com/openMVG/openMVG/blob/4e78ed73a575beae88199fc440183a75e03068c5/src/cmakeFindModules/OpenMVGConfig.cmake.in#L87-L93

I can hack around this by patching `OpenMVGConfig.cmake.in` to use the `OpenMP_FOUND` variable instead:
```cmake
set(OpenMVG_USE_OPENMP "@OpenMP_FOUND@")
```

The latter fix seems simpler, but I'm not sure if it's appropriate given the rest of the configuration system.

For posterity, I'm running on macOS 15.3 on an M3 Max MBP.

csparker247:
This fixes the issue where the installed OpenMVG CMake package was expecting to find OpenMP even though OpenMP wasn't found during the library build.

It looks like the code for disabling the OpenMP configuration was incorrectly tied to the `OpenMVG_USE_OPENMP==FALSE` condition instead of `OpenMP_FOUND`.

Closes #2393


src/CMakeLists.txt
```diff
@@ -29,6 +29,9 @@ endif()
 
 set(CMAKE_FIND_FRAMEWORK LAST)
 
+# Helper utility for overriding cached variables
+include(UpdateCacheVariable)
```
csparker247: Moved this up here because this command is used in more than just the OpenMP block.

src/CMakeLists.txt
```diff
@@ -178,12 +181,9 @@ if (OpenMVG_USE_OPENMP)
...
     endif()
     list(APPEND OPENMVG_LIBRARY_DEPENDENCIES OpenMP::OpenMP_CXX)
+  else (OPENMP_FOUND)
+      UPDATE_CACHE_VARIABLE(OpenMVG_USE_OPENMP OFF)
   endif (OPENMP_FOUND)
-else (OpenMVG_USE_OPENMP)
-    option(OpenMVG_USE_OPENMP "Use OpenMP for parallelization" OFF)
-    include(UpdateCacheVariable)
-    UPDATE_CACHE_VARIABLE(OpenMVG_USE_OPENMP OFF)
-    remove_definitions(-DOPENMVG_USE_OPENMP)
```
csparker247, line 186: This definition never gets added if OpenMP isn't found, so no need to remove it.

pmoulon:
Thank you for your contribution @csparker247

PR #2424, opened 2025-08-29
Update dependencies which use old CMake version minimums

### LINKED ISSUE #2422: Included dependencies break on CMake 4.0+
As of CMake 4.0, compatibility with versions of CMake older than 3.5 is removed [[link]](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html#policy-version). I found out the hard way that many of this project's included dependencies and third_party sources still use old versions for `cmake_minimum_required`, thus breaking OpenMVG builds on CMake 4.0+. While there is a [workaround](https://cmake.org/cmake/help/latest/variable/CMAKE_POLICY_VERSION_MINIMUM.html), it seems like a good time to update/fix these dependencies.

I'll slowly work through identifying what and how each can be updated. Thus far, I've identified that the following projects have issues (possibly more to come):
- `dependencies/osi_clp` - Seems to just require an updated `cmake_minimum_required` call, but maybe we want to update the source code at the same time?
- `third_party/lemon` - Seems like an update `cmake_minimum_required` is all that's needed, and there's not really new code upstream.
- `third_party/ceres-solver` - The latest version of Ceres is 2.2.0, which requires CMake 3.16+, so it avoids this issue. However, there have been many changes to the existing Ceres source inside OpenMVG to make it play nicely with the OpenMVG build system. I've attempted to update to v2.2.0 and patch these changes back in, but not sure if I'm missing anything.
- `rerun` - Rerun is actually okay, but the internal `arrow` library has the CMake error, even in the newest versions.

csparker247:
## Summary

Update bundled and internal dependencies to resolve CMake 4.1.0+ incompatibilities, where CMake 4.0 removed support for `cmake_minimum_required` versions < 3.5.

### OSI-CLP
- Submodule updated (ref: csparker247/osi_clp) with `cmake_minimum_required` raised to 3.10
- See upstream fork PR: openMVG-thirdparty/osi_clp#7

### LEMON
- Updated to latest upstream tip
- `cmake_minimum_required` raised to 3.10
- Fixed version-string handling

### Ceres-solver
- Updated to v2.2.0
- Built library target renamed to `openMVG_ceres` with aliases `OpenMVG::ceres` and `ceres_static`
- Target export and uninstall target disabled to keep it internal-only

### GLFW
- Submodule re-pointed from stale `elmindreda` fork to official `glfw/glfw` repository
- Updated to 3.4 release (declares `cmake_minimum_required(VERSION 3.4...3.28)`, resolving its own CMake 4 incompatibility)
- OpenGL examples (main_VO, sfmViewer) verified building against new API

### libpng
- Bundled version updated: 1.6.6 → 1.6.58 (drop-in source replacement)
- Resolves macOS build break where 1.6.6 unconditionally included non-existent Classic Mac OS `<fp.h>`
- ARM NEON optimizations disabled (`PNG_ARM_NEON_OPT=0`) to avoid references to unavailable bundled symbols; insignificant for fallback-only usage

### RERUN
- Reference updated to v0.24.1
- Note: does not resolve underlying issue and can be reverted

## CMake 4.x Compatibility Fixes

- **src/CMakeLists.txt**: minimum version raised from 3.9 to range form `3.10...4.3`
- **Bundled jpeg and png CMakeLists**: `cmake_minimum_required` updated to range form (`2.8...4.3` and `2.4.4...4.3` respectively) to prevent hard errors under CMake 4 while preserving legacy policy behavior
- **src/cmakeFindModules/OptimizeForArchitecture.cmake**: replaced removed `exec_program` command with `execute_process` to comply with CMP0153 policy

## Build Status

Verified locally on macOS/arm64 with CMake 4.3.4:
- Full clean build completes successfully with all targets linking and running
- Internal Ceres 2.2.0, bundled png 1.6.58 + jpeg, and GLFW 3.4 OpenGL viewers functional
- Image-I/O unit test passes

## Considered but deferred

The dependency-freshness items in #2441 were evaluated for inclusion here and deliberately left out. This PR is scoped to "make the build work under CMake 4"; the items below are API-bearing or out-of-repo and belong in their own PRs so this one stays reviewable and low-risk.

- **cereal → 1.3.2**: Not a CMake 4 blocker — cereal is a header-only `INTERFACE` target, so its `CMakeLists.txt` is never processed. Bumping it (current fork stalls ~1.2.2 → `USCiLab/cereal@v1.3.2`) is an API-bearing serialization change that warrants its own PR with round-trip compatibility testing.
- **OSI 0.108.12 / CLP 1.17.11**: Cannot be done by editing this repo — it requires re-vendoring inside the `openMVG-thirdparty/osi_clp` submodule first, then bumping the submodule ref here. This PR includes a ref to the currently opened PR openMVG-thirdparty/osi_clp/pull/7 which simply bumps the CMake minimum required version to 3.10.
- **Eigen 5**: API-bearing and larger in scope — Eigen 5 requires C++14, while openMVG currently builds with `-std=c++11`. Adopting it means a project-wide C++ standard bump and compile sweep, out of scope here.

Also noted (not in #2441): `pixi.toml` still pins `cmake 3.28.1`, so the pixi CI lane does not exercise CMake 4.

## References

Fixes #2422 (CMake 4 build failure); relates to #2441 (dependency version freshness)


csparker247:
My plan here is to have one PR where I just get everything running. If we think it's simple enough, then it can be merged as is, but if it's too complicated for a single PR, then I'll split each dependency into its own PR.

csparker247:
This is now a pretty comprehensive set of CMakeList changes and version bumps. 

Given the number of things that are basically just revendored, I might advocate for no longer vendoring as many of these as we can. But this is complete, nonetheless.

PR #2431, opened 2025-12-12
fix rotation magnitude calculation

mitjap:
Looking at this code I expect that rotation magnitude should be the angle from identity. 
https://github.com/openMVG/openMVG/blob/c92ed1be7c25068c0cdc0927e876a9ec92e97ce8/src/openMVG/sfm/pipelines/global/GlobalSfM_rotation_averaging.cpp#L196-L199

Current implementation of `getRotationMagnitude` uses some custom logic. I think the correct magnitude should be the same as converting the rotation matrix to angle-axis representation and take the angle. Especially since this value is compare to some angular threshold. 

My implementation follows this equation: 
image
https://en.wikipedia.org/wiki/Rotation_matrix#Determining_the_angle

### DIFF
--- a/src/openMVG/numeric/numeric.cpp
+++ b/src/openMVG/numeric/numeric.cpp
@@ -56,8 +56,7 @@ Mat3 RotationAroundZ(double angle) {
 }
 
 double getRotationMagnitude(const Mat3 & R2) {
-  const Mat3 R1 = Mat3::Identity();
-  double cos_theta = (R1.array() * R2.array()).sum() / 3.0;
+  double cos_theta = (R2.trace() - 1.0) / 2.0;
   cos_theta = clamp(cos_theta, -1.0, 1.0);
   return std::acos(cos_theta);
 }
--- a/src/openMVG/numeric/numeric.h
+++ b/src/openMVG/numeric/numeric.h
@@ -138,12 +138,13 @@ inline double R2D( double radian )
 }
 
 /**
-* @brief Compute mean rotation magnitude of the given rotation matrix
-* @param R2 Input rotation matrix
-* @return magnitude of the rotation (in radian)
-* @note Assuming R2 is a correct rotation matrix
-* @note Mean is computed using the matrix column dot products to an Identity matrix
-*/
+ * @brief Compute the rotation magnitute of the given rotation matrix
+ * @param R2 Input rotation matrix
+ * @return magnitude of the rotation (in radian)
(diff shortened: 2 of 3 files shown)


pmoulon:
My approach was closer to a `Frobenius similarity`.
Did you see any difference running it in practice?

mitjap:
There wasn't any noticeable difference. 

I visualized both methods for easier understanding.
Both methods calculate the magnitude from `R2`'s trace (`x = R2.trace()`).

https://www.desmos.com/calculator/gwimnp60ze
image

`f(x)` is current implementation
`g(x)` is new implementation
`h(x)` is relative difference between the two

The relative difference seems to be more or less constant, `~0.8`. To account for this difference we could adjust the threshold. Current threshold is `5.0f`. Adjusted threshold would be `~6.1f`.

fallenmi (approved):
No blocking findings.

I checked exact head `e65892027c09e8d2bdbcaff8d9ec499d15495f45` and GitHub's current synthetic merge `da3c81e1c8620479a0241da597baae2ec03b2e72` into `develop` `c76d87244fb3590fb8b9a752be34f07411057ae2`. For a proper rotation matrix, `trace(R) = 1 + 2 cos(theta)`, so the new clamped `(trace - 1) / 2` formula returns the canonical SO(3) angle that the function and all four callers describe.

An independent oracle compiled against the repository's actual base, head, and merge implementations covers the 5-degree filter boundary and rotations through 360 degrees. The old formula reports 5.001 degrees as 4.082867 degrees and would accept it under the documented `< 5.0` threshold; head/merge report 5.001 degrees and reject it. Across the sweep, base reaches 70.528779 degrees of error at 180 degrees, while head and merge match the canonical angle within `1e-7` degrees.

Verification: the submitted numeric test passes, all five tests in the owning numeric directory pass on exact head and the current synthetic merge, and `git diff --check` is clean.

CI caveat: the Actions rollup is stale and red from 2025-12-12. Those jobs ended within seconds with no recorded steps; the retained annotations contain only exit code 129 and the logs have expired (HTTP 410). CodeFactor's only reported issue is the minor blank line at the start of the new test body. Neither exposes a functional blocker, and the exact local head/merge verification above passes.

AI disclosure: OpenAI Codex was used to inspect the diff and all callers, construct and run the exact base/head/merge oracle and tests, and draft this review. The cited commands and outputs were run against the exact SHAs above.

PR #2432, opened 2025-12-12
fix triplet validation in translation averaging

mitjap:
Validation is incorrect. This fixes it. 

I was expecting big reduction of noisy triplets to be removed but there is a high probably they were already removed by rotation averaging. This fix does have a small influence on final calculation but not that much.


pmoulon:
Thank you, it seems indeed I was too fast on this

PR #2433, opened 2025-12-12
fix submodule initalization in github workflows

mitjap:
This PR fixes incorrect initialization of git submodules in workflows. Some workflows now complete successfully.


pmoulon:
Thank you @mitjap , we may have to fix other minor things

PR #2436, opened 2026-01-26
Implement sparse eigen-solver for rotation averaging in L2 method

yanxke:
Introduce a sparse eigen-solver for rotation averaging, enhancing scalability for large image sets while maintaining a fallback to a dense solver for smaller image sets.

I ran into a wall when trying process 4500+ images and this step took hours.  Now it's very fast.  The threshold (2048) can probably be tweaked further but for now it works.

### DIFF
--- a/src/openMVG/multiview/rotation_averaging_l2.cpp
+++ b/src/openMVG/multiview/rotation_averaging_l2.cpp
@@ -20,6 +20,8 @@
 #endif
 
 #include "openMVG/system/logger.hpp"
+#include "third_party/spectra/include/Spectra/MatOp/SparseSymMatProd.h"
+#include "third_party/spectra/include/Spectra/SymEigsSolver.h"
 
 //--
 //-- Implementation related to rotation averaging.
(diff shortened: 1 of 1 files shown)


pmoulon:
Thanks

fallenmi (changes requested):
The sparse path cannot recover the three nullspace eigenvectors on a valid connected graph at the first matrix size that selects it.

At exact head `90e8fc0eaef034269729f4ee7f82adc795bbd3cc`, I exercised the repository's actual `L2RotationAveraging` with 683 cameras and identity relative rotations on an overlapping-triangle strip: edges `(i,i+1)` plus `(i,i+2)`. The graph is connected, every edge belongs to a zero-composition-error triangle, and no edge is a bridge, so it meets the triplet-composition and largest-bi-edge-connected prefilters used by the global pipeline. Exact base `c76d87244fb3590fb8b9a752be34f07411057ae2` returns `true` with 683 output rotations. Head forms a 2049 x 2049 `A^T A`, selects `SymEigsSolver` with `k=3`, `ncv=20`, and `SmallestMagn`, returns `NotConverging` after its default iteration budget, and `L2RotationAveraging` returns `false` with zero outputs. This reproduced 10/10. The same head oracle with 682 cameras forms a 2046 x 2046 matrix, uses the dense path, and succeeds, so adding one camera flips a valid input from success to failure solely at the 2048 boundary.

The bundled Spectra documentation notes that the standard solver may fail for eigenvalues near zero and provides shift-and-invert for this case. Please use an eigensolver/configuration appropriate for the three-dimensional nullspace (or a validated fallback) and add a greater-than-2048 regression using a sparse, prefilter-valid graph.

Verification: built exact GitHub merge `5d6b72300c920b17180ecff9d77a79564cb507a8`; all 7 existing owning tests pass, but their L2 cases remain on the dense branch. I also ran exact base/head actual-function oracles and an independent raw reproduction against the bundled Spectra implementation.

AI disclosure: This review was prepared with Codex assistance, including diff inspection and exact-revision build/test/oracle execution. The evidence above is included to make the result independently reproducible.

